Issue #217, opened 2026-08-11
Date item fields still discard a configured adornment


phmatray:
Noted in #189 (PR #196) and acknowledged in the #184 release note.

## Problem

Date item fields still **discard a configured adornment**. `CollectionFieldComponent.RenderDateField` passes `rendersAdornment: false` on purpose: `MudDatePicker` defaults to `Adornment.End` with its own calendar icon, and forwarding a field's (usually unset) adornment would silently strip that icon on every date item field.

So `.WithAdornment(...)` on a date item field is accepted and dropped — the same class of silent-discard that #184, #191 and #192 each closed elsewhere.

## Proposed solution

Forward the adornment only when one is actually configured, leaving MudDatePicker's calendar icon in place otherwise. That distinction needs `GetItemFieldAdornment` to tell "unset" from "explicitly set to None", which the current resolution collapses.

Related: the `ShrinkLabel` diagnostic deliberately passes `null` for this path so a configured-but-dropped adornment is not reported — that pairing needs revisiting together with this fix.

## Area

`FormCraft.ForMudBlazor/Features/CollectionField/CollectionFieldComponent.razor.cs`.

---

Follow-up from #196 (issue #189).

phmatray:
Subsumed by #203 — blocked on it rather than fixed on its own.

Bar applied: *would resolving #203 resolve this too?* Yes. The `rendersAdornment: false` in `CollectionFieldComponent.RenderDateField` disappears with the method itself. Once date item fields render through the same date component as standalone ones, the `MudDatePicker` calendar-icon trade-off is decided **once**, in one place, instead of being a divergence between two.

**Blocked, not closed**, deliberately: #203's Assumptions allow fallback approach **B**, which keeps two renderers. Under B this remains a live divergence needing its own fix. Re-open the decision if #203 is re-scoped.

Issue #214, opened 2026-08-11
Nothing enforces the RenderTreeBuilder sequence-number budget in CollectionFieldComponent


phmatray:
Found while working #189 (PR #196), and confirmed the hard way while merging #193/#194/#195/#196 — the budget had to be re-verified by hand to know the four PRs could land together.

## Problem

The `RenderTreeBuilder` sequence-number budget in `CollectionFieldComponent` is enforced **only by a comment**.

`AddCommonFieldAttributes` has grown 8 → 13 attributes across #177/#181/#184/#190/#192. `CallerAttributeStart` is 20. At 20 the common block would collide with the caller block and **corrupt render-tree diffing silently** — no exception, no failed assertion, just attributes overwriting each other.

Current allocation:

| Block | Constant | Used |
|---|---|---|
| common | `CommonAttributeStart = 1` | 1–13 |
| caller | `CallerAttributeStart = 20` | 20–23 |
| text | `TextAttributeStart = 30` | 30–33 |

No test or analyzer catches an overflow. The comment on `CommonAttributeStart` states the constraint but nothing checks it.

## Proposed solution

A test that renders an item field of each type and asserts the highest sequence number each block emits stays below the next block's start. Deriving the starts from the measured counts, rather than hard-coding 20/30, would make the constraint self-maintaining.

## Area

`FormCraft.ForMudBlazor/Features/CollectionField/CollectionFieldComponent.razor.cs` + tests.

---

Follow-up from #196 (issue #189).

phmatray:
Subsumed by #203 — blocked on it rather than fixed on its own.

Bar applied: *would resolving #203 resolve this too?* Yes, and more completely than a guard would. #203 Task 3 removes `AddCommonFieldAttributes`, the `CommonAttributeStart`/`CallerAttributeStart` constants and the hand-rolled `RenderTreeBuilder` block outright. There is then no sequence-number budget left to enforce, so a test enforcing one would be testing deleted code.

**Blocked, not closed**, deliberately: #203's Assumptions allow fallback approach **B**, which keeps the imperative renderer. Under B the budget stays real and this test becomes worth writing. Re-open the decision if #203 is re-scoped.

Issue #213, opened 2026-08-11
ShrinkLabelDiagnosticCollector keys conflicts by bare FieldName, so item fields collide with top-level ones


phmatray:
Found while working #189 (PR #196).

## Problem

`ShrinkLabelDiagnosticCollector` keys conflicts by **bare `FieldName`**. A top-level field and a collection item field that share a name overwrite each other, undercounting the aggregated warning.

## Proposed solution

Key by the qualified identifier. `GetItemFieldIdentifier` in `CollectionFieldComponent` already builds the `Items[0].Name` form — use that as the key so the two namespaces cannot collide.

## Area

`FormCraft.ForMudBlazor` — diagnostics.

---

Follow-up from #196 (issue #189).

## 🛠️ Implementation plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** two fields that share a name but live in different scopes are counted separately by the aggregated ShrinkLabel warning.

**Architecture:** qualify the collector key on the collection path with the owning collection's name. The collector itself is unchanged — its contract is already "the key is the field's identity"; the bug is that the item path passed an identity that is only unique within one item form.

**Tech Stack:** .NET 8/10, Blazor, MudBlazor 9, xUnit + bUnit + Shouldly (`MudBlazorTestBase`).

### Global Constraints

- **Default branch is `dev`.** PRs target `dev` (`--base dev`).
- **Merge style: squash.** PR title `fix(mudblazor): <subject> (#213)` — enforced by `pr-title-lint.yml`.
- **Commit identity:** `git -c user.email=phmatray@gmail.com -c user.name="Philippe Matray"`.
- ⛔ **The warning must stay once per FIELD, not once per row.** A 50-row collection must not produce 50 entries — that is what `_warnedItemFields` protects and it must keep working.
- ⛔ **Do not key by label.** The collector's own docs explain why: two fields can legitimately share a label ("Name" in Billing and in Shipping), and keying on it would merge them — the same undercount, moved.
- `dotnet test --filter` is inert (MTP0001); run the full suite. `TreatWarningsAsErrors=true` is deliberate.

### A second collision this issue does not name

The issue describes a top-level field colliding with an item field. The same key is also shared by
**two different collections** in one form: `Orders`' item field `Name` and `Extras`' item field
[...]

Issue #190, opened 2026-08-11
Collection item fields emit the HTML5 Required attribute, contradicting the project's validation convention


phmatray:
## Problem / motivation

The two render paths disagree about the HTML5 `Required` attribute, and the collection path is the
one that breaks the project's stated convention.

`CLAUDE.md` records the rule explicitly:

> `Required()` adds validation but NOT HTML5 required attribute · Browser validation disabled via
> `novalidate` · All validation through FluentValidation · **MudBlazor components don't include
> `Required` attribute**

Measured with a probe rendering the same `.Required("…")` field two ways:

| Render path | `MudTextField.Required` |
|---|---|
| Ordinary field | `False` ✅ — matches the convention |
| Inside `.WithItemForm(...)` | **`True`** ❌ |

`CollectionFieldComponent.AddCommonFieldAttributes` emits `builder.AddAttribute(…, "Required",
field.IsRequired)`. No component-path renderer emits `Required` at all — `grep -r "Required"
FormCraft.ForMudBlazor/**/*.razor` returns nothing.

**Consequence.** Inside a collection, MudBlazor's own required validation runs *alongside*
FluentValidation, so a blank required item field can produce two error messages with different
wording for one problem — and the message MudBlazor generates is not the one the developer wrote in
`.Required("Product name is required")`.

Found while reviewing #188, which added a render-path parity test. `Required` was initially in that
test's compared set but was never configured by the test, so it asserted nothing; removing it
surfaced this genuine divergence, which the test now documents rather than hides.

## Proposed solution

Stop emitting `Required` from `AddCommonFieldAttributes`, matching the component path and the
documented convention, then move `Required` into the parity test's compared set with the shared
`Configure` delegate actually calling `.Required(...)`.

Validation behaviour is unaffected: `Required()` still registers its validator, and FluentValidation
still produces the developer's message. What disappears is MudBlazor's *second*, differently-worded
browser-style error.

## Alternatives considered

- **Emit `Required` on the component path too**, making both paths agree in the other direction.
  Rejected: it contradicts the documented convention, re-enables browser-flavoured validation the
  project deliberately disables with `novalidate`, and would change behaviour on every existing form
  rather than only on collection item fields.
- **Leave the divergence and document it.** Rejected — the parity test would have to carry a permanent
  exception, and duplicate error messages are a real user-visible defect.
- **Make it configurable.** Disproportionate for an attribute the project has already decided about.

## Area

[...]

PR #250, opened 2026-08-12
refactor(mudblazor): render collection item fields through IFieldRendererService (#203)

### LINKED ISSUE #203: Converge the two render paths: collection item fields duplicate the field renderers instead of using them
## Problem / motivation

The same defect has now been found and fixed **four times, reactively, years apart**:

| Issue | Attribute the collection path got wrong |
|---|---|
| #146 | `Variant` |
| #177 | `ShrinkLabel` |
| #184 | `Adornment` / `AdornmentIcon` / `AdornmentColor` |
| #190 | `Required` (this time by doing *too much*, not too little) |

Each was fixed the same way: add one more attribute to
`CollectionFieldComponent.AddCommonFieldAttributes`. None addressed why the two paths can disagree at
all.

**The root cause is duplication.** Ordinary fields render through the declarative path — a `.razor`
component per field type, selected by `IFieldRendererService`. Fields inside `.WithItemForm(...)` render
through an imperative path: `CollectionFieldComponent` builds the tree by hand with a
`RenderTreeBuilder`. Two implementations of "render a field", so any presentation behaviour added to one
is silently absent from the other.

The tell is in the file itself. `CollectionFieldComponent.razor:4` declares:

```razor
@inject IFieldRendererService FieldRendererService
```

…and **never calls it**. The component already has the service that would render fields correctly, and
hand-rolls the work instead.

**This is not finished.** `RenderPipelineParityTests`' divergence list still names the queue:

- the raw `"Required"` attribute (collection-only, opened by #193)
- `InputType`, `Lines`, `MaxLength`, `Autocomplete` (#189)
- `OnAdornmentClick` (an explicit non-goal of #184)
- `EnablePasswordToggle`, `Mask` (per #189's work in progress)
- plus `RenderBooleanField`, which bypasses `AddCommonFieldAttributes` entirely, so **every** shared
  attribute is absent for `bool` item fields

At one bug report per attribute, the current approach has a known, unbounded tail.

## Proposed solution

Make the collection item path **use** `IFieldRendererService` instead of duplicating it, so item fields
render through the same components as ordinary fields and inherit every present and future presentation
attribute for free.

The parity test then stops being a list of known divergences and becomes a near-tautology — which is the
point.

This is a refactor with **no intended behaviour change**, and that has to be provable: the existing
suites (`CollectionAdornmentTests`, `CollectionShrinkLabelTests`, `CollectionRequiredTests`,
`CollectionFieldEditContextTests`, `RenderPipelineParityTests`) are the contract and must pass
unmodified, with the deliberate exception of the divergences that *disappear*.

## Alternatives considered

- **Keep adding attributes to `AddCommonFieldAttributes`.** Status quo. Cheap per bug, unbounded in
  total, and each fix ships an inconsistency window. It also cannot fix `RenderBooleanField`, which
  takes none of them.
- **Extract a shared "presentation attribute" descriptor consumed by both paths.** A middle option:
  keeps two renderers but derives their attributes from one source. Less invasive, and it fixes the
  drift class without touching the render architecture — but leaves two code paths for value binding,
  validation wiring and event handling, which have their own divergence potential.
- **Delete the collection item path and require custom templates.** Rejected — a large, user-visible
  capability regression.

## Area

FormCraft.ForMudBlazor — collection field rendering, render pipeline

---

**Follow-up from #193.** **Related:** #146, #177, #184, #190, #189, #191, #192

<details>

### Problem / context

`CLAUDE.md` already names this as a standing hazard:

> **Two render paths exist and must agree.** Ordinary fields go through the *declarative* path …
> fields inside `.WithItemForm(...)` go through the *imperative* path … `RenderPipelineParityTests`
> exists to keep them honest — presentation attributes have repeatedly drifted between them.

The parity test (added in #188) was the right response to a symptom: it makes drift *visible*. It does
not make drift *impossible*, and its own doc comment carries a growing list of attributes it declines to
compare because the paths genuinely disagree.

Worth being honest about why the imperative path exists: it renders item fields against a *nested* model
(`TItem`) while wiring validation to nested `FieldIdentifier`s on the *parent* (`Items[0].ProductName`,
per #91). That's real complexity the declarative path doesn't currently express — it is the reason this
refactor is a project rather than a patch.

### Approaches

**A. Route item fields through `IFieldRendererService`.**
*Pros:* removes the duplication at its root; every existing and future presentation attribute applies to
both paths automatically; deletes a large hand-rolled `RenderTreeBuilder` block including its
sequence-number hazard; `RenderBooleanField`'s gaps disappear with it. *Cons:* the biggest change; must
preserve nested `FieldIdentifier` wiring, reorder behaviour and per-item value binding exactly.

**B. Share a presentation-attribute descriptor between the two renderers.**
*Pros:* much smaller and lower-risk; kills the specific recurring bug class (presentation drift).
*Cons:* leaves two renderers, so binding/validation/event drift stays possible; the parity list shrinks
but does not disappear.

**C. Status quo + a stricter parity test.**
*Pros:* nearly free. *Cons:* still one bug report per attribute; the test can only compare what someone
remembers to configure — which is precisely how #190's `Required` sat in the list asserting nothing.

### Recommendation

**A**, staged, with **B** as an acceptable first milestone if A proves too large in one go. The
sequencing matters more than the choice: land the parity guarantees *before* removing the old path, so
the refactor is provably behaviour-preserving rather than hopefully so.

Suggest doing this **after** the open per-attribute issues (#189, #191, #192) land, so the refactor
inherits their tests as its safety net rather than racing them.

</details>

<details>

### Goal

One implementation of "render a field". A field configured identically renders identically whether or
not it sits inside `.WithItemForm(...)`, by construction rather than by vigilance.

### Scope

- Render collection item fields through `IFieldRendererService` and the per-type `.razor` components.
- Preserve nested `FieldIdentifier` wiring (`Items[0].ProductName`), per-item value binding, reorder
  behaviour, and the `ShrinkLabel` diagnostic.
- Remove `AddCommonFieldAttributes`, its sequence-number block, and `RenderBooleanField`'s bespoke
  attribute handling once nothing calls them.
- Reduce `RenderPipelineParityTests`' divergence list to the entries that remain genuinely intentional.

### Non-goals

- Any change to the `FormCraft` core (it stays UI-framework-agnostic).
- New presentation features. This is behaviour-preserving; new attributes belong in their own issues.
- Changing the public builder API.

### Behaviour

```mermaid
flowchart TD
    subgraph today
      A1["ordinary field"] --> B1["IFieldRendererService<br/>→ .razor component"]
      A2["item field"] --> B2["CollectionFieldComponent<br/>RenderTreeBuilder by hand"]
      B1 -.->|drift| B2
    end
    subgraph after
      C1["ordinary field"] --> D["IFieldRendererService<br/>→ .razor component"]
      C2["item field"] --> D
    end
```

### Key files

- `FormCraft.ForMudBlazor/Features/CollectionField/CollectionFieldComponent.razor{,.cs}` — the imperative path.
- `FormCraft.ForMudBlazor/Fields/**` — the component path renderers.
- `FormCraft/Rendering/IFieldRendererService.cs`, `FieldRendererService.cs` — the selector.
- `FormCraft.ForMudBlazor.UnitTests/Components/RenderPipelineParityTests.cs` — the guard.

### Validation rules

- Every existing collection test passes **unmodified**, except where a documented divergence disappears.
- Nested validation messages still attach to `Items[i].FieldName` and clear on correction (#91).
- Reordering a mixed item form still moves values with their rows (the keyless-loop hazard from #184).
- `bool` item fields gain the shared presentation attributes rather than silently ignoring them.
- The parity test's divergence list shrinks; any surviving entry is intentional and stated.

### Edge cases

- **Nested `FieldIdentifier`s** are the crux: item components must notify the *parent* `EditContext`.
- **Date fields** keep MudBlazor's own calendar adornment — the component path must not erase it (#184).
- **Reorder + keyless loops**: differing frame counts per row already caused a positional-pairing hazard.
- **The `ShrinkLabel` diagnostic** must judge what the path actually renders, not what was configured (#183).
- **`.WithAttribute("Required", true)`** currently works only on the collection path; converging makes it
  work on both, which is a behaviour change to state deliberately (see the typed-builder follow-up).

### Assumptions

- The declarative path can be given a nested-model context without redesigning `IFieldRendererService`.
  If it cannot, approach **B** is the fallback and this issue should be re-scoped rather than forced.
- Base branch is `dev`. This is a minor-version refactor, not a patch.

</details>

## 🛠️ Implementation plan

**Goal:** collection item fields render through `IFieldRendererService`, eliminating the duplicate render
path that has produced four separate bugs.

**Architecture:** all changes in `FormCraft.ForMudBlazor`; `FormCraft` core untouched. Behaviour-preserving
— the existing suites are the contract.

**Tech stack:** .NET 8 / 10 multi-target, Blazor, MudBlazor 9, xUnit + bUnit + Shouldly (`MudBlazorTestBase`).

**Global constraints:**
- `TreatWarningsAsErrors=true` — the build fails on any warning.
- `dotnet test --filter` is silently ignored here (`MTP0001`); run the whole suite.
- No MudBlazor type may leak into `FormCraft` core.
- Existing collection tests must pass **unmodified** — editing one to make the refactor pass is the
  failure mode to avoid. If a test must change, that is a behaviour change and needs stating.
- Land after #189, #191 and #192 so their tests act as the safety net.

### Task 1: Characterisation tests — pin today's behaviour before changing it

**Files:** create `FormCraft.ForMudBlazor.UnitTests/Components/CollectionRenderCharacterisationTests.cs`.

**Interfaces:** none — tests only.

- [x] **Step 1:** Write tests capturing item-field behaviour the current suites do not yet pin: per-item value binding across add/remove/reorder, nested `FieldIdentifier` notification, and the rendered attribute set for each of text / numeric / date / bool.
- [x] **Step 2:** Run `dotnet test -c Release` → PASS against the CURRENT implementation (these describe today, not the target).
- [x] **Step 3:** Commit: `test(mudblazor): characterise collection item rendering before converging paths`.

### Task 2: Give the declarative path a nested-model rendering context

**Files:** modify `FormCraft.ForMudBlazor/Fields/**` shared base as needed; modify `FormCraft/Rendering/` context types if required; test `CollectionRenderCharacterisationTests.cs`.

**Interfaces:** a render context able to bind a field against `TItem` while reporting validation against the parent model's `Items[i].FieldName`.

- [x] **Step 1:** Write a failing test rendering a single item field through `IFieldRendererService` with a nested context, asserting it binds to the item and notifies the parent `EditContext`.
- [x] **Step 2:** Run the suite → FAIL.
- [x] **Step 3:** Implement the nested context, keeping `FormCraft` core free of MudBlazor types.
- [x] **Step 4:** Run the suite → PASS.
- [x] **Step 5:** Commit: `feat(mudblazor): render fields against a nested item context`.

### Task 3: Route item fields through the service and delete the hand-rolled path

**Files:** modify `FormCraft.ForMudBlazor/Features/CollectionField/CollectionFieldComponent.razor{,.cs}`.

**Interfaces:** `RenderItemField` delegates to `FieldRendererService`; `AddCommonFieldAttributes`, `RenderBooleanField`'s bespoke attributes and the sequence-number constants are removed.

- [x] **Step 1:** Switch `RenderItemField` to call the injected `FieldRendererService` (the service the component already injects and never uses).
- [x] **Step 2:** Run `dotnet test -c Release` → expect failures; fix the **implementation**, never the existing tests.
- [x] **Step 3:** Delete `AddCommonFieldAttributes`, the `CommonAttributeStart`/`CallerAttributeStart` constants, and the now-dead per-type render methods.
- [x] **Step 4:** Run the suite → PASS with `CollectionAdornmentTests`, `CollectionShrinkLabelTests`, `CollectionRequiredTests` and `CollectionFieldEditContextTests` **unmodified**.
- [x] **Step 5:** Verify `bool` item fields now honour the shared presentation attributes, and update `CollectionAdornmentTests`' inertness test — that one SHOULD change, and the change is the point.
- [x] **Step 6:** Commit: `refactor(mudblazor): render collection item fields through IFieldRendererService`.

### Task 4: Collapse the divergence list and release-note the convergence

**Files:** modify `FormCraft.ForMudBlazor.UnitTests/Components/RenderPipelineParityTests.cs`; modify `README.md`.

**Interfaces:** none new.

- [x] **Step 1:** Move every attribute the two paths now share into `Presentation()`, configuring each in the shared `Configure` delegate, and delete the divergence bullets that no longer apply.
- [x] **Step 2:** Run the suite → PASS. Verify the guard bites by breaking one attribute deliberately, then restore.
- [x] **Step 3:** Add a release-note bullet under the README's `## 🎉 Unreleased` section covering the behaviour changes users will actually see — notably `bool` item fields gaining presentation attributes, and `.WithAttribute("Required", true)` becoming path-independent.
- [x] **Step 4:** Run `dotnet build -c Release` and `dotnet test -c Release` → both green.
- [x] **Step 5:** Commit: `test(mudblazor): compare the full presentation set across both render paths`.

phmatray:
Implements #203.

Closes #203.

Collection item fields render through a hand-rolled `RenderTreeBuilder` path instead of the
`IFieldRendererService` the component already injects and never calls. That duplication has produced
four separate bugs (#146, #177, #184, #190) and a growing divergence list. This converges the two
paths so item fields inherit every present and future presentation attribute for free.

code-review pass.

### Plan
- [x] Task 1: Characterisation tests — pin today's behaviour before changing it
- [x] Task 2: Give the declarative path a nested-model rendering context
- [x] Task 3: Route item fields through the service and delete the hand-rolled path
- [x] Task 4: Collapse the divergence list and release-note the convergence

### What changed

`CollectionFieldComponent.razor.cs` drops from 836 lines to ~215: the hand-rolled type switch,
`AddCommonFieldAttributes`, `AddTextInputAttributes`, the four per-type `Render*Field` methods, the
`CommonAttributeStart`/`CallerAttributeStart`/`TextAttributeStart` index space and every
`GetItemField*` resolver are gone. `RenderItemField` is now a single call to
`IFieldRendererService.RenderField`, passing the **item** as the model.

What the item cannot carry is supplied by a cascaded `CollectionItemFieldScope`: the owning
collection's name (so form-wide diagnostics keep #213's `Items[].Field` qualification) and a
once-per-field latch (so a 50-row collection does not emit 50 identical warnings, which the old path
latched by hand). Validation still reports to the parent `EditContext` under `Items[i].Field` (#91).

`MudBlazorDateTimeFieldComponent` had to learn #217's adornment binding, or date **item** fields
would have silently lost it in the move — which also fixes the mirror-image bug on ordinary date
fields, where `.WithAdornment(...)` was being dropped. Both release-noted.

`SequenceNumberBudgetTests` (#214) is deleted: it guarded the sequence-number index space of the
renderer this PR removes. The hazard no longer exists rather than being merely guarded.

### Follow-ups

- **`.claude/skills/repo-profile.md`'s *Worktree home* note is stale.** It warns that
  `.claude/worktrees/` is not git-ignored because two `.gitignore` entries were collapsed onto one
  line — that was fixed in #223 and `dev` ignores the directory correctly. The note still describes
  the broken state (and `main` has not been updated either). Not touched here: unrelated to this
  change.
- **`Mask` is still an unimplemented stub.** `MudBlazorTextFieldComponent.GetMask()` returns `null`
  and nothing calls it; FormCraft stores a `string` where MudBlazor wants an `IMask`. Before #203
  this was one of two remaining parity-list entries; it is now a **shared gap** — both placements
  render no mask and can no longer differ — so it should either be implemented or the builder
  surface removed.


phmatray:
**Follow-up disposition** (triaged at merge):

- **`ValueExpression` recompiled every render** → filed as #269 (`type:refactor`, `priority:medium`). Sharpened while writing it up: `Context.CurrentValue` is read only by `FieldComponentBase`'s *fallback* branch (`FieldComponentBase.cs:99`), so for every field whose property resolves the compile result is computed and **discarded** — the work is not merely expensive, it is usually unused.
- **`Mask` unimplemented stub** → **already resolved.** #211 landed on `dev` (`8484dba`) while this PR was being merged and implements it on both placements via `TextMaskMap`. No issue filed; the parity list entry was updated to `✅ Resolved in #211` during the merge.
- **Stale *Worktree home* note in `repo-profile.md`** → folded into #264, which already owns the `.gitignore` root cause, rather than filed separately. That comment also reports a measurement worth acting on: the defect #264 describes is **already fixed on `dev`** (`.claude/worktrees/` and `test-results/` both `check-ignore` clean), so #264 may be closeable — what remains is `main` and the stale doc.

Issue #225, opened 2026-08-11
Preserve the MTP test-results artifact on failing CI runs


phmatray:
## Problem / motivation

**A test that fails only on CI cannot currently be diagnosed from CI.**

FormCraft's suites run on **Microsoft.Testing.Platform**, which prints only a summary line to stdout
and writes every per-assertion detail — the failing test names, the Shouldly messages, the stack
traces — to `FormCraft.UnitTests/bin/<cfg>/<tfm>/TestResults/*.log`. That file is the *only* place the
detail exists.

Neither CI workflow preserves it on failure:

| Workflow | Uploads test-results? | On failure? |
|---|---|---|
| `.github/workflows/ci.yml` (`build-and-test`) | **no artifact upload at all** | — |
| `.github/workflows/continuous.yml` (`ubuntu-latest`) | yes, `Publish: test-results` | **no** — the step has no `if: always()`, so a failed `Run: Continuous` skips it |
| `.github/workflows/release-please.yml` (`nupkg`) | yes | **yes** — already carries `if: always()` |

So when `./build.cmd Test` fails, the step that would have saved the evidence is skipped and the log
dies with the runner. `gh run view <id> --log` yields the summary line and nothing else:

```
Failed! - Failed: 3, Passed: 752, Skipped: 1, Total: 756 - FormCraft.UnitTests.dll (net10.0|x64)
error run failed: Tests failed: '…/TestResults/FormCraft.UnitTests_net10.0_x64.log'
```

— a path to a file that no longer exists, and **no test names**.

**Measured cost.** This is not hypothetical: during #200 a CI-red/local-green failure took several
rounds of hypothesis-and-check to identify, because `gh run download` returned *"no valid artifacts
found"* for the failing run. The cause was eventually found by comparing **total test counts** between
CI and local (756 vs 757) and inferring that the compiled source differed — an inference that would
have been a five-second read of three test names had the log survived.

This is also already recorded as a known repository gotcha ("MTP writes assertion detail to
`TestResults/*.log`, which the workflow does not publish: an only-on-CI test failure is undiagnosable
remotely as-is"), so it is a documented trap that has now been paid for at least once.

## Proposed solution

Preserve the evidence on the path that matters — failure — by copying the pattern
`release-please.yml` **already uses**:

1. Add `if: always()` to `continuous.yml`'s `Publish: test-results` step.
2. Add a test-results upload to `ci.yml`, likewise `if: always()`. `ci.yml` is the workflow that runs
[...]

PR #252, opened 2026-08-12
ci: emit test reports natively under Microsoft.Testing.Platform (#231)

### LINKED ISSUE #231: The build's test-reporting wiring is inert under MTP, and release-please.yml still uploads the dead path
## Problem / motivation

Follow-up from #225 (PR #230). While proving that a red CI run now preserves its assertion log, the
underlying reporting wiring turned out to be **entirely inert**, and one workflow still carries the
dead path #225 fixed in the other two.

`build/Build.cs`'s `Test` target asks for a results directory and two loggers:

```csharp
.SetResultsDirectory(TestResultsDirectory)          // <repo>/test-results
.SetLoggers("trx", $"html;LogFileName={TestResultsDirectory / "test-results.html"}")
```

`dotnet test` forwards those as the VSTest properties `VSTestResultsDirectory` / `VSTestLogger`, and
Microsoft.Testing.Platform **ignores both**:

```
warning MTP0001: VSTest-specific properties are set but will be ignored when using
Microsoft.Testing.Platform. The following properties are set: VSTestLogger; VSTestResultsDirectory;
```

Consequences, all measured rather than inferred:

1. **`test-results/` is never created.** `./build.sh Test` succeeds and leaves no such directory.
2. **No trx or html report has ever been produced**, so the `Test` target's
   `.Produces(TestResultsDirectory / "*.trx")` / `"*.xml"` contract describes files that do not exist.
3. **`.github/workflows/release-please.yml`'s `Publish: test-results` step still points only at
   `test-results`.** It has `if: always()`, so it runs — and uploads nothing. #225 deliberately did not
   touch that file (its plan's Global Constraints forbade it), so it is the one remaining copy of the
   defect.
4. Every `./build.cmd Test` emits two `MTP0001` warnings, which is noise that trains readers to ignore
   warnings in a repo whose whole build runs under `TreatWarningsAsErrors`.

None of this breaks a build, which is precisely why it survived: the wiring *reads* correct and
produces nothing.

## Proposed solution

Make the build's stated test-reporting contract true, or delete it — do not leave a third state where
it is declared and inert.

- **Emit reports natively.** Microsoft.Testing.Platform has its own options (`--report-trx`,
  `--report-trx-filename`, `--results-directory`). Drive those instead of the VSTest properties, so
  `test-results/` is genuinely populated and the `.Produces(...)` contract holds.
- **Or drop `SetResultsDirectory`/`SetLoggers` and the two `.Produces(...)` lines**, and rely solely on
  the `**/TestResults/*.log` upload #225 introduced. Smaller, and honest.
- **Either way, fix `release-please.yml`'s `path:`** to match the two workflows #225 corrected.
- **Pin it.** Add a guard test in the style of `FormCraft.UnitTests/Ci/TrustedPublishingWorkflowTests.cs`
  asserting that all three `Publish: test-results` steps carry `if: always()` and a path that can
  actually match. Nothing currently stops any of this regressing.

## Alternatives considered

- **Leave it.** The upload from #225 already delivers the diagnostic, so this is not urgent — but it
  leaves a build target that claims outputs it never produces, and one workflow whose artifact is
  silently empty. That is the exact shape of trap #225 was filed about.
- **Only fix `release-please.yml`.** Removes the third dead path but keeps the inert `Build.cs` wiring
  and the recurring `MTP0001` warnings.

## Area

Build/CI infrastructure — `build/Build.cs`, `.github/workflows/release-please.yml`, and a new guard
test in `FormCraft.UnitTests/Ci/`. No library code.

**Related:** #225, #230. Note the read recipe recorded in `ci.yml`: the MTP log is UTF-16LE with ANSI
escapes, so a plain `grep` on it matches nothing and looks like an empty artifact.

## 🛠️ Implementation plan

> **Seeded by `implement-issue`.** This issue was filed without a plan, so the checklist below was
> derived from its *Proposed solution* and pinned to measurements taken on `dev` (see *Findings*).

### Findings that decide the fork

The issue offers two shapes — emit reports natively, or delete the wiring. **Emit natively wins**, and
it costs nothing:

- `xunit.v3` 3.2.2's Microsoft.Testing.Platform runner already ships the reporters. `--help` on the
  built test executable lists `--results-directory`, `--report-xunit-trx[-filename]`,
  `--report-xunit-html[-filename]`, `--report-xunit`, `--report-junit`, `--report-nunit`,
  `--report-ctrf`. **No new package reference is required.**
- Measured: `dotnet test FormCraft.sln -c Release --no-build -- --results-directory <root>/test-results
  --report-xunit-trx --report-xunit-html` populates `test-results/` with one `.trx` and one `.html`
  **per test assembly**, default-named `<user>_<machine>_<timestamp>` so the two assemblies cannot
  collide.
- **`--results-directory` also relocates the MTP `.log` diagnostics** — the artifact #225 exists to
  preserve — out of `<project>/bin/<cfg>/<tfm>/TestResults/` and into `test-results/`. Verified: after
  such a run, `find . -path "*/bin/*/TestResults/*" -name "*.log"` returns **0** files and
  `test-results/` holds both `*_net10.0_arm64.log` files.
- That last point is what makes `release-please.yml` correct *for free*: its `path: test-results` stops
  being a dead path the moment `Build.cs` fills the directory. Conversely the `**/TestResults/*.log`
  glob in `ci.yml`/`continuous.yml` becomes the dead one, and the long comment blocks in both files
  asserting "`test-results/` is never created" become false.
- `.Produces(TestResultsDirectory / "*.xml")` is unsatisfiable either way — the HTML reporter writes
  `.html` and the xunit XML reporter writes `.xunit`. Nothing emits `.xml`.

### Global Constraints

- **Base branch is `dev`, not `main`.** The PR targets `dev`; `main` is the release branch.
- `TreatWarningsAsErrors=true` is deliberate — never relax it to make a build pass.
- **No library code.** Only `build/Build.cs`, `.github/workflows/*.yml`, and a new test in
  `FormCraft.UnitTests/Ci/`.
- `dotnet test --filter` is inert under MTP (MTP0001) — always run the **full** suite

### Task 1: Drive MTP's native report options from `Build.cs`

- [x] **Step 1:** Add `FormCraft.UnitTests/Ci/TestReportingTests.cs`, in the style of
      `TrustedPublishingWorkflowTests`, asserting that `Build.cs` no longer sets the VSTest logger /
      results-directory properties MTP ignores, and that it does drive the MTP report options. Run
      `dotnet test -c Release` and confirm **RED**.
- [x] **Step 2:** In `build/Build.cs`'s `Test` target, drop `.SetResultsDirectory(...)` and
      `.SetLoggers(...)` and pass `--results-directory`, `--report-xunit-trx`, `--report-xunit-html`
      through to the test application instead.
- [x] **Step 3:** Correct the `Test` target's `.Produces(...)` contract to name the files that are
      actually emitted, replacing the unsatisfiable `*.xml`.
- [x] **Step 4:** Run `./build.sh Test`; confirm `test-results/` holds a `.trx`, an `.html` and a
      `.log` per test assembly, and that the run emits **no `MTP0001`** warning.
- [x] **Step 5:** Run `dotnet test -c Release`; confirm **GREEN**.
- [x] **Step 6:** Commit `build: emit test reports through Microsoft.Testing.Platform (#231)`

### Task 2: Make all three `Publish: test-results` uploads honest

- [x] **Step 1:** Extend `TestReportingTests` with assertions that **every** workflow step named
      `Publish: test-results` carries `if: always()`, points at a path the build now fills, and
      tolerates a missing path. Run `dotnet test -c Release` and confirm **RED** on
      `release-please.yml`.
- [x] **Step 2:** Bring `release-please.yml`'s `Publish: test-results` step to the same shape as the
      one in `ci.yml` and `continuous.yml`.
- [x] **Step 3:** Rewrite the now-false comment blocks in `ci.yml` and `continuous.yml` that state
      `test-results/` is never created, and retire the superseded `**/TestResults/*.log` glob.
- [x] **Step 4:** Run `dotnet test -c Release`; confirm **GREEN**.
- [x] **Step 5:** Commit `ci: align every test-results upload with the directory the build fills (#231)`

### Task 3: Prove and record the failure path

- [x] **Step 1:** Add a temporary always-failing test and run `./build.sh Test`, so the red path — the
      only one worth preserving, per #225 — is measured rather than assumed.
- [x] **Step 2:** Confirm the failing run still leaves the per-assembly `.log` carrying the assertion
      detail **and** a `.trx` recording the failure in `test-results/`; then remove the temporary test.
- [x] **Step 3:** Update `ci.yml`'s "Reading it" recipe so the documented `gh run download` /
      `iconv` / `grep` steps match the artifact's measured layout and encoding.
- [x] **Step 4:** Run `dotnet test -c Release`; confirm **GREEN**.
- [x] **Step 5:** Commit `docs(ci): record the measured layout of the test-results artifact (#231)`

phmatray:
Implements #231.

Closes #231.

The issue was filed without a `🛠️ Implementation plan`, so one was derived from its *Proposed
solution*, pinned to measurements taken on `dev`, and posted to the issue before any code was
written. The plan resolves the issue's own fork — *emit natively* vs *delete the wiring* — in favour
of **emitting natively**, because `xunit.v3`'s Microsoft.Testing.Platform runner already ships the
reporters (`--report-xunit-trx`, `--report-xunit-html`, `--results-directory`), so the honest fix
costs no new dependency.

### Plan
- [x] Task 1: Drive MTP's native report options from `Build.cs`
- [x] Task 2: Make all three `Publish: test-results` uploads honest
- [x] Task 3: Prove and record the failure path

### What changed

- **`build/Build.cs`** — the `Test` target drops `.SetResultsDirectory()`/`.SetLoggers()` (forwarded
  as `VSTestResultsDirectory`/`VSTestLogger`, which MTP discards with `MTP0001`) and passes MTP's own
  `--results-directory`, `--report-xunit-trx`, `--report-xunit-html` instead. `.Produces(...)` now
  names `*.trx`, `*.html`, `*.log` — the files actually written — rather than the unsatisfiable
  `*.xml`. The directory is cleaned first, since timestamped report names accumulate rather than
  overwrite, and a red run followed by a green one would otherwise publish stale failures.
- **All three workflows** — one path, `test-results`, plus `if-no-files-found: ignore`.
  `release-please.yml` (the copy #225 was forbidden to touch) was already pointing there and became
  correct without moving; `ci.yml`/`continuous.yml` shed the `**/TestResults/*.log` glob, whose
  premise no longer holds. The long comment blocks stating `test-results/` is never created were
  rewritten, since that is now false.
- **`FormCraft.UnitTests/Ci/TestReportingTests.cs`** — 8 guard facts. The workflow set is *derived*
  (any workflow invoking `./build.cmd Test|Pack|Continuous`) rather than listed, so a fourth workflow
  is held to the contract on the day it lands — which is exactly how `release-please.yml` was left
  behind. `.Produces(...)` extensions are cross-checked against the report flags, so the assertion
  cannot be satisfied by copying a literal.

### The finding that made this cheap

`--results-directory` does not only place the trx/html — it **relocates MTP's per-assembly diagnostic
log** (the artifact #225 exists to preserve) out of `<project>/bin/<cfg>/<tfm>/TestResults/` and into
`test-results/`. Measured: after such a run, `find . -path "*/bin/*/TestResults/*" -name "*.log"`
returns 0 files. That is why fixing the *emitting* end also fixed `release-please.yml`'s "dead path"
without editing its `path:` at all.

### Verification

- `dotnet build -c Release` — 0 warnings, 0 errors (the repo's real gate: `TreatWarningsAsErrors`).
- `dotnet test -c Release` — 788 + 354 pass, 0 fail.
- `./build.sh Test` — green, **no `MTP0001`**, and `test-results/` holds a `.trx`, `.html` and `.log`
  per test assembly.
- **Red path measured, not assumed.** A temporary always-failing test was added and
  `./build.sh Test` run: `test-results/` still received both logs and both trx, the log carried the
  full Shouldly assertion detail and stack trace, and the trx recorded `outcome="Failed"` under the
  test's fully-qualified name. The probe was then removed.
- **Proven in real CI.** The `build-and-test` run on `80b6c03` produced a `test-results` artifact
  containing 2 logs + 2 trx + 2 html from the runner — the first non-empty one this repo has ever
  produced. The `Counters` read `total="789" failed="0"` and `total="354" failed="0"`.
- The guard tests were **mutation-tested**: deleting `if: always()` from `continuous.yml` turns
  `Every_TestResults_Upload_Should_Run_On_Failure` red, and only it.

### A second, unrelated silent failure — fixed

The log-reading recipe `ci.yml` documents produced **nothing** when run against a real artifact. The
artifact holds one log per test assembly, and a single `iconv` over both leaves the second file's BOM
inline immediately before its first line — which is the line naming the first failure — so
`grep '^failed '` cannot match it. Byte-level confirmation: `357 273 277` (`EF BB BF`) directly
preceding `failed`. Since the first file's failures would print but the second's never would, this
failed in the same silent, plausible-looking way as the wiring this issue is about. The recipe is now
a per-file loop, verified against both a locally reddened run and the real CI artifact.

### Code review

A review pass raised 15 findings. Ten were real and are fixed in `f60b791`:

- **The recovery recipe truncated its own output.** `grep -A5` stops five lines after the `failed`
  line, which is inside Shouldly's difference table — so the `at ...` stack frames never printed,
  while the next comment line claimed they did. Now `-A30`, verified to reach the stack trace.
- **The BOM mechanism is now stated exactly.** BSD grep drops a UTF-8 BOM only at the *start* of the
  stream and GNU grep drops none, so `^failed ` misses a failing assembly's first line on Linux
  always, and on macOS as soon as more than one file is piped in. The recipe now strips the BOM
  explicitly instead of relying on either behaviour.
- **`if-no-files-found: ignore` could hide a repeat of this very defect** — an empty `test-results/`
  would upload silently. Fixed at the source rather than in the workflows: the `Test` target now
  `Assert.NotEmpty`s the `*.trx` it just claimed to produce. Verified by mutation — removing
  `--report-xunit-trx` fails the build with a message naming the cause, instead of going green.
- **The "backed by a reporter" check matched prefixes**, so `--report-xunit-html-filename` alone
  satisfied `--report-xunit-html` (names a file, enables nothing) and `--report-xunit-trx` satisfied
  `--report-xunit`. Now word-boundary matched; mutation-tested.
- **Workflow discovery was too narrow** — it missed `run: |` blocks, `./build.sh`, and the
  `Publish`/`PublishIfNeeded`/`Release` targets that also reach `Test` via `DependsOn`.
- **One assertion searched raw text**, so a comment quoting the step name would have satisfied it.
- Plus: the `*.log` mapping now describes the real mechanism (`dotnet test`'s MSBuild integration,
  not MTP's opt-in `--diagnostic`), workflow files are read once instead of ~20 times, the `path:`
  check is pinned to the whole value, and an unresolvable `<see cref="DotNetTest"/>` became `<c>`.

Three findings I disagreed with, having measured them:

- *"The log carries no ANSI escapes, so the `sed` is unnecessary."* True locally, **false in CI** —
  which is the only place this recipe is ever used. The downloaded artifact contains
  `ESC[33mskipped ESC[m…`; MTP colourises the status word on the runner, so an uncoloured `^failed `
  would not match. The `sed` stays, and the comment now says where the escapes come from.
- *"Concatenating the logs still matches every `^failed ` line."* Not for this artifact: the failing
  assembly's log opens *with* its `failed <test>` line, so the mid-stream BOM lands directly on it —
  `EF BB BF` immediately before `failed`, measured. Concatenated `grep -c '^failed '` returns 0 where
  the per-file loop returns 1. The reviewer's log evidently began with the runner banner.
- *"Cleaning `test-results/` before the run destroys the failing run's reports."* A real cost, but
  the alternative — unbounded accumulation under timestamped names — makes `.Produces(*.trx)`
  ambiguous about which run it describes, and `Clean` already means "empty this". Kept, with the
  tradeoff written into the comment.

Two more were fair but are deferred as follow-ups below (helper duplication, per-assembly report
names); one — that the workflow assertions are file-scoped rather than job-scoped — is now recorded
as a known limitation in the test class, since closing it needs a real YAML parse.

## Follow-ups

- **`TestReportingTests` and `TrustedPublishingWorkflowTests` now carry near-duplicate helpers**
  (`LocateRepoRoot`, `WithoutComments`, workflow-step extraction). They should share one
  `Ci/` helper. Deliberately not done here: folding the older class onto a new shared type would put
  an unrelated refactor of a passing guard suite into a build/CI fix.
- **The trx and html reports carry no assembly identity.** The two `.log` files are named
  `FormCraft.UnitTests_net10.0_x64.log` / `FormCraft.ForMudBlazor.UnitTests_…`, but the trx and html
  are `runner_<machine>_<timestamp>.trx` — distinguishable only by a microsecond. Attribution is
  recoverable from the file's contents, but not from its name. The fix is a per-project
  `--results-directory`, which means invoking `DotNetTest` per test project instead of once for the
  solution — a larger change to how the suite runs than this issue warrants.
- **The workflow assertions are file-scoped, not job-scoped.** A workflow that ran the build in one
  job and uploaded from another would satisfy them while uploading nothing, since the second job gets
  a fresh workspace. Closing it needs a real YAML parse; noted in the test class for now.
- **`.worktrees/` is not in `.gitignore`** (`.claude/worktrees/` is). Only matters to agent tooling
  that uses that second location, but a worktree left there is staged by any `git add -A`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,11 +39,19 @@ jobs:
       # `if-no-files-found: ignore` covers the one legitimate miss: a run that fails in Compile,
       # before any test executes and therefore before `test-results/` exists.
       #
-      # Reading it: the log is UTF-16LE and carries ANSI colour escapes, so a plain `grep '^failed '`
-      # matches nothing and looks exactly like an empty artifact. Use:
+      # Reading it (measured against a deliberately reddened run, #231). The log is UTF-16LE and
+      # carries ANSI colour escapes, so a plain `grep '^failed '` matches nothing and looks exactly
+      # like an empty artifact. And it must be ONE iconv PER FILE: the artifact holds one log per
+      # test assembly, and concatenating them leaves the second file's BOM inline immediately before
+      # its first line — which is the line naming the first failure — so `^failed ` cannot match it
+      # either. Both traps fail the same silent way, which is why the loop is spelled out:
       #   gh run download <run-id> -n test-results -D /tmp/tr
-      #   iconv -f UTF-16LE -t UTF-8 /tmp/tr/*.log | sed $'s/\\x1b\\[[0-9;]*m//g' \
-      #     | grep -A5 '^failed '
+      #   for f in /tmp/tr/*.log; do
+      #     iconv -f UTF-16LE -t UTF-8 "$f" | sed $'s/\\x1b\\[[0-9;]*m//g' | grep -A5 '^failed '
+      #   done
+      # That prints the failing test names with their Shouldly detail and stack traces. The same
+      # run's *.trx carries each failure as outcome="Failed" under its fully-qualified testName, and
+      # the *.html is the same report rendered for a browser.
       - name: 'Publish: test-results'
         if: always()
         uses: actions/upload-artifact@v7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,19 +39,32 @@ jobs:
       # `if-no-files-found: ignore` covers the one legitimate miss: a run that fails in Compile,
       # before any test executes and therefore before `test-results/` exists.
       #
-      # Reading it (measured against a deliberately reddened run, #231). The log is UTF-16LE and
-      # carries ANSI colour escapes, so a plain `grep '^failed '` matches nothing and looks exactly
-      # like an empty artifact. And it must be ONE iconv PER FILE: the artifact holds one log per
-      # test assembly, and concatenating them leaves the second file's BOM inline immediately before
-      # its first line — which is the line naming the first failure — so `^failed ` cannot match it
-      # either. Both traps fail the same silent way, which is why the loop is spelled out:
+      # Reading it. Every clause below was measured against a deliberately reddened run and against a
+      # real downloaded artifact (#231) — the naive form of this command silently prints nothing,
+      # which is indistinguishable from "the suite passed".
+      #
+      #  - The log is UTF-16LE, one file per test assembly, each starting with a BOM.
+      #  - It carries ANSI colour escapes ON CI and not locally: MTP colourises the status word, so a
+      #    failure line arrives as ESC[31mfailed ESC[m<test>. Measured on a downloaded artifact,
+      #    which is the only copy this recipe is ever pointed at. Hence the escape strip.
+      #  - The BOM has to be stripped explicitly, because grep implementations disagree about it:
+      #    BSD grep drops one at the START of the stream, GNU grep drops none. A failing assembly's
+      #    log can open with its `failed <test>` line, so an unstripped BOM makes `^failed ` miss the
+      #    first failure — on Linux always, and on macOS as soon as more than one file is piped in.
+      #  - `-A30`, not `-A5`: Shouldly's difference table alone runs past twenty lines, so a smaller
+      #    window cuts off the `at ...` stack frames — the part naming the offending line.
+      #
       #   gh run download <run-id> -n test-results -D /tmp/tr
       #   for f in /tmp/tr/*.log; do
-      #     iconv -f UTF-16LE -t UTF-8 "$f" | sed $'s/\\x1b\\[[0-9;]*m//g' | grep -A5 '^failed '
+      #     echo "== $f"
+      #     iconv -f UTF-16LE -t UTF-8 "$f" \
+      #       | sed $'s/\\xef\\xbb\\xbf//g; s/\\x1b\\[[0-9;]*m//g' \
+      #       | grep -A30 '^failed '
       #   done
-      # That prints the failing test names with their Shouldly detail and stack traces. The same
-      # run's *.trx carries each failure as outcome="Failed" under its fully-qualified testName, and
-      # the *.html is the same report rendered for a browser.
+      #
+      # That prints each failing test with its Shouldly detail and stack trace. The same run's *.trx
+      # carries each failure as outcome="Failed" under its fully-qualified testName, and the *.html
+      # is the same report rendered for a browser.
       - name: 'Publish: test-results'
         if: always()
         uses: actions/upload-artifact@v7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,34 +18,36 @@ jobs:
           DOTNET_CLI_TELEMETRY_OPTOUT: 1
 
       # Microsoft.Testing.Platform prints only a summary line to stdout and writes every per-assertion
-      # detail — failing test names, Shouldly messages, stack traces — to
-      # <project>/bin/<cfg>/<tfm>/TestResults/*.log. Without this upload a red run here is
-      # undiagnosable remotely: `gh run view --log` yields a count and a path to a file that died with
-      # the runner. That cost was paid for real during #200. (#225)
+      # detail — failing test names, Shouldly messages, stack traces — to a per-assembly `.log`.
+      # Without this upload a red run here is undiagnosable remotely: `gh run view --log` yields a
+      # count and a path to a file that died with the runner. That cost was paid for real during
+      # #200. (#225)
       #
       # `if: always()` is the entire point — the failure path is the only one worth preserving, and it
       # is exactly the path a bare step skips.
       #
-      # ⚠️ The path deliberately does NOT rely on Nuke's `test-results/` directory. Build.cs asks for it
-      # via DotNetTest's --results-directory/--logger, but `dotnet test` passes those as the VSTest
-      # properties VSTestResultsDirectory/VSTestLogger, which Microsoft.Testing.Platform ignores
-      # outright (warning MTP0001). So `test-results/` is never created, and an upload pointed only at
-      # it produces an empty artifact that merely looks like a fix. Measured, not assumed:
-      # continuous.yml has carried such a step for some time and no run of it has ever produced a
-      # test-results artifact. `test-results` is still listed so the artifact gains the trx/html the
-      # day the build actually emits them.
+      # One path, and as of #231 it is a real one. Build.cs used to ask for `test-results/` through
+      # DotNetTest's VSTest surface, which `dotnet test` forwards as VSTestResultsDirectory /
+      # VSTestLogger and Microsoft.Testing.Platform ignores outright (warning MTP0001) — so the
+      # directory was never created, no trx or html was ever written, and this step leaned on a
+      # `**/TestResults/*.log` glob pointing into the projects' bin directories instead. Build.cs now
+      # passes MTP's own `--results-directory`, which puts the trx and html reports AND that
+      # per-assembly log under `test-results/`. The old glob is deliberately gone rather than kept as
+      # a fallback: every workflow that tests reaches Nuke's Test target, so it could now only ever
+      # match nothing — the declared-but-inert shape #231 was filed about.
+      #
+      # `if-no-files-found: ignore` covers the one legitimate miss: a run that fails in Compile,
+      # before any test executes and therefore before `test-results/` exists.
       #
       # Reading it: the log is UTF-16LE and carries ANSI colour escapes, so a plain `grep '^failed '`
       # matches nothing and looks exactly like an empty artifact. Use:
       #   gh run download <run-id> -n test-results -D /tmp/tr
-      #   iconv -f UTF-16LE -t UTF-8 /tmp/tr/**/TestResults/*.log | sed $'s/\\x1b\\[[0-9;]*m//g' \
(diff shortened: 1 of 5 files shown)


phmatray:
Recording one finding from this PR that is worth knowing but not worth a tracked issue.

**`.worktrees/` is not covered by `.gitignore`.** `.gitignore` ignores `.claude/worktrees/` but not the
sibling `.worktrees/` convention some agent tooling uses. A worktree left in an unignored directory is
staged by any `git add -A` as a single `160000` gitlink pointing at a commit no clone can fetch — not a
diff that stands out in review. One line fixes it if anyone cares; noted here rather than filed because
it concerns agent tooling rather than the library.

For contrast, `main`'s copy of `.gitignore` currently collapses two entries onto one line
(`/test-results/.claude/worktrees/`), so on that branch neither `/test-results/` nor `.claude/worktrees/`
is ignored. `dev` already has them on separate lines, so the next release merge carries the fix — worth
confirming it survives rather than being resolved back to the collapsed form.

PR #268, opened 2026-08-12
feat(demo): redesign the demo site around the code-to-form binding

phmatray:
## Why

The demo site was stock MudBlazor — Roboto, Material Blue, a gradient stats band, six identical feature cards — and its hero split FormCraft's one interesting idea across three separate boxes (*Configure Fields* / *Live Preview* / *View Generated Code*).

FormCraft's whole product is a **transform**: a typed C# builder chain becomes a real, validated form. Nothing on the page showed that as a single object. This makes it the hero.

## The binding

`BindingHero` puts the builder chain beside the form it produced. Both halves derive from one field list, so they cannot disagree, and the right half is a **real `FormCraftComponent`** — not a picture of one.

- Hovering a chain line lights the control it created (and the reverse).
- Adding a field writes the line and renders the control in the same frame.
- Green marks the pairing on both sides at once — the tie *is* the colour.

## Identity

| | |
|---|---|
| **Colour** | The .NET platform violet `#512BD4` carries structure. Green `#14A06B` has exactly one job — "this is bound", "this is valid" — and never decorates. |
| **Type** | Bricolage Grotesque (display), IBM Plex Sans (body), IBM Plex Mono promoted to a first-class role because code *is* the subject matter here. |
| **Tokens** | `wwwroot/css/tokens.css`; `FormCraftTheme.cs` mirrors them for MudBlazor. Change one, change the other. |

## Navigation and UX

- **Command palette** on <kbd>⌘K</kbd> / <kbd>Ctrl+K</kbd> / <kbd>/</kbd>, ranking all 27 pages by title → concepts → description. The site had no search at all.
- **Theme persistence** honouring `prefers-color-scheme`, resolved before first paint so returning visitors get no flash of the wrong theme.
- The six feature cards and the four-tab showcase collapse into one capability table that routes into the demos.

## Bugs found along the way

- `prism.min.css` ships Prism's **light** theme. On the dark code ground every documentation page rendered black text on near-black — effectively unreadable. Replaced with the site's own code palette, at a specificity that does not depend on `<link>` order.
- MudBlazor's `button:focus { outline: none }` (0,1,1) outranked a bare `:focus-visible` rule (0,1,0), leaving buttons with **no visible keyboard focus**.
- Difficulty was green/amber/red, which reads as *status* — "Advanced" in red looked like something had gone wrong. Replaced with `LevelMeter`, a filled 1–3 scale.
- The page claimed "600+ tests" and "18 demos". Actual: **1,134 passing** and **20 demos**. The spec strip now states verified facts and uses a lower bound for the test count so it will not go stale.

Also adds a skip link, and deletes `InteractivePlayground`, `AnimatedCounter` and `FeatureShowcase` (all now unreferenced) along with the home-page CSS they used.

## Scope

Demo app only. No change to `FormCraft` or `FormCraft.ForMudBlazor`, the README, or the markdown documentation *sources* — only how they render.

## Verification

- `dotnet build /p:TreatWarningsAsErrors=true` — clean.
- `dotnet test` — 780 + 354 passing, 1 skipped, 0 failed (unchanged from base).
- Checked by hand in Chrome at 1440px and 390px, in light and dark: the binding highlight, field add/remove, the palette (open, filter, rank, keyboard hints), theme persistence across reload, skip link as first tab stop, and focus rings on MudBlazor and custom controls.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


phmatray:
Recording one environment observation from this PR's run — not actionable enough to be its own issue, but worth having written down somewhere retrievable.

**The worktree this branch was built in got retargeted mid-session by something outside the job.** Its reflog:

```
checkout: moving from review-252 to pr254
checkout: moving from worktree-groovy-wondering-bee to review-252
```

Neither switch came from this session. The branch under `.claude/worktrees/groovy-wondering-bee` therefore drifted from the one the job started on to `pr254` — an unrelated accessibility PR — while work was in progress.

That was caught before committing, and the work was moved onto a branch cut from this PR's base (`5634649` on `dev`) after confirming `pr254`'s own commits touch **zero** demo-app files, so nothing here is layered on top of it. Build and tests were re-verified on the new base before the PR was opened.

Flagging it because the next job may not notice: had the retarget landed a few minutes later, this PR's commit would have gone onto `pr254`'s branch and quietly widened someone else's PR.

Issue #266, opened 2026-08-12
Warn when a mask silently blanks a stored value that does not fit the pattern


phmatray:
## Problem / motivation

Follow-up from #211 (PR #253). A masked field whose **stored value does not fit the pattern renders empty**, while the model quietly keeps the original:

```csharp
var model = new Contact { Phone = "N/A" };      // legacy data, or a value entered before the mask existed

.AddField(x => x.Phone, field => field
    .WithLabel("Phone")
    .WithAttribute("Mask", "(000) 000-0000"))
```

The user sees a **blank** input, submits the form without touching it, and `"N/A"` survives untouched. Nothing is logged, nothing is thrown, and the form looks like it simply has an empty optional field.

The mechanism is MudBlazor's, not FormCraft's: on the first parameter pass `MudBaseInput` builds the display text by running the value through the mask (`updateValue: false`), the mask rejects the non-conforming characters, `Text` collapses to empty — and because `updateValue` is false, `Value` is never written back. Measured and pinned by `TextField_With_A_Mask_Should_Blank_An_Existing_Value_That_Does_Not_Fit`.

**This is not new MudBlazor behaviour — it is newly *reachable*.** Until #211 the `Mask` attribute did nothing at all, so no FormCraft field could hit it. Turning masking on turns this on with it, which makes it an upgrade hazard for exactly the forms most likely to have legacy data: the ones that already wrote `.WithAttribute("Mask", …)` and got nothing for it.

The README's Unreleased section warns about it in prose. Prose is not a signal a developer receives while debugging "why is this field empty".

## Proposed solution

Warn, once per field, when a field's initial value is non-blank and the configured mask renders it as blank. This is the same shape as the diagnostics this library already ships, and it slots into the same family:

| Diagnostic | Warns when |
|---|---|
| `MaskedLinesDiagnostic` (#207) | a multi-line setting is dropped so a password stays masked |
| `PasswordAdornmentDiagnostic` (#219) | the visibility toggle displaces a configured adornment |
| `ShrinkLabelDiagnosticCollector` (#181) | `ShrinkLabel=false` is overridden by a placeholder or start adornment |
| **this** | a mask blanks a stored value that does not fit it |

Emit on both render paths, latched per field name so a 50-row collection re-rendering on every keystroke produces one warning rather than fifty.

## Alternatives considered

[...]

PR #273, opened 2026-08-12
feat(mudblazor): add a typed WithMask builder for text field masks (#265)

### LINKED ISSUE #265: Add a typed WithMask builder instead of the "Mask" magic string
## Problem / motivation

Follow-up from #211 (PR #253), which made `Mask` work on both render paths. It works, but the only way to reach it is a magic string:

```csharp
.WithAttribute("Mask", "(000) 000-0000")
```

That is the same defect #204 closed for the native-required opt-in — "add a typed builder method instead of a magic string" — and it costs three things here rather than one:

1. **No discoverability or type safety.** `"Mask"` is spelled by the caller. A typo is silently inert, which is precisely the failure #211 existed to fix; #253 named the key as `TextMaskMap.AttributeName` internally, but the *caller* still types the literal.
2. **`CleanDelimiters` is unreachable, so the model always stores the delimited text.** MudBlazor's `PatternMask.CleanDelimiters` decides whether `GetCleanText()` strips the literals; it defaults to `false` and FormCraft never sets it. So a field masked `(000) 000-0000` writes `"(555) 123-4567"` to the model, never `"5551234567"`. Storage, APIs and validators keyed to raw input see punctuation, and there is no knob to change it. Documented as a behaviour change in the README's Unreleased section and pinned by `TextField_With_A_Mask_Should_Write_The_Masked_Text_To_The_Model`.
3. **A pre-built `IMask` is accepted and silently discarded.** `.WithAttribute("Mask", new RegexMask(…))` is the natural thing for a MudBlazor user to write — it compiles, builds and renders. Both paths read the attribute as `string?` (`GetAttribute<string?>` / `GetItemFieldAttribute<string?>`), whose `value is T typed` test fails for an `IMask` and falls back to `null`. So `RegexMask`, `BlockMask` and `MultiMask` are unreachable, and asking for one does nothing without a warning — the "looks supported, does nothing" shape this repo keeps closing (#184, #191, #192, #211).

## Proposed solution

Add a typed `.WithMask(...)` builder to `MudBlazorFieldBuilderExtensions`, alongside `.WithAdornment(...)`, `.WithNativeRequired(...)` and `.WithAutocomplete(...)`, with two overloads:

```csharp
.WithMask("(000) 000-0000")                          // pattern, delimiters kept (today's behaviour)
.WithMask("(000) 000-0000", cleanDelimiters: true)   // model receives 5551234567
.WithMask(new RegexMask(@"^\d{4}$"))                 // any MudBlazor IMask
```

The builder writes named attribute constants; `TextMaskMap.Resolve` grows to read them, so both render paths keep resolving through the one method #211 introduced and cannot drift. `.WithAttribute("Mask", "…")` keeps working — it is the documented surface today and this issue is additive.

`ITextFieldComponent.Mask` in the core project stays `string?` and stays UI-agnostic: the typed builder lives entirely in `FormCraft.ForMudBlazor`, which is where MudBlazor's `IMask` belongs.

## Alternatives considered

**Widen `ITextFieldComponent.Mask` to `object?` (or an abstraction).** Rejected. It is a breaking change to the core contract, and it drags a UI-framework concept into the framework-agnostic project — the invariant CLAUDE.md and the repo profile both state. #253's code review flagged an earlier revision for documenting MudBlazor's pattern syntax in that very file.

**Add a second magic-string attribute, `"MaskCleanDelimiters"`.** Rejected. It reaches the smallest of the three problems while making the largest one worse — two undiscoverable strings to spell correctly instead of one, which is the situation #204 was filed about.

**Resolve the mask at build time and store an `IMask` in the configuration.** Rejected. Field configuration is built once and shared by every rendered instance, while a `BaseMask` carries the live `Text`/`CaretPos` of one input — storing a resolved mask would hand the same stateful object to every row of a collection. #253 chose per-render construction deliberately for this reason.

## Area

`FormCraft.ForMudBlazor` — `MudBlazorFieldBuilderExtensions`, `TextMaskMap`, and the two render paths (`MudBlazorTextFieldComponent`, `CollectionFieldComponent`).

**Related:** #211, #204, #203

<details>

### Problem / context

#211 closed the "mask does nothing" bug by writing the missing string→`IMask` conversion and binding it on both render paths. What it deliberately did not do is widen the *surface*: the issue's scope was parity, so `.WithAttribute("Mask", <pattern string>)` remained the only entry point, and #253's own follow-up list named the two gaps that leaves.

The repo has already litigated this exact shape once. #193 made `.WithAttribute("Required", true)` the supported opt-in; #204 then replaced it with `.WithNativeRequired()` because a magic string is undiscoverable and mistypes silently. Every comparable MudBlazor-specific setting here now has a typed builder — `.WithAdornment`, `.WithAutocomplete`, `.AsPassword`, `.WithNativeRequired`. `Mask` is the outlier.

### Approaches

**A. Typed `.WithMask(...)` with a pattern overload and an `IMask` overload.** Two attribute constants (pattern + options) plus one for a supplied instance; `TextMaskMap.Resolve` reads them. Discoverable, type-safe, covers all three gaps, additive. Cost: `Resolve`'s signature grows beyond a single string, and the `IMask` overload needs a documented answer on instance sharing.

**B. Pattern overload only; leave `IMask` unreachable.** Smaller. Closes the discoverability and `CleanDelimiters` gaps and leaves `RegexMask`/`BlockMask`/`MultiMask` unavailable — but the silent-discard is then *still* live, so the "looks supported, does nothing" bug survives in the exact library this issue is about.

**C. Do nothing; document `.WithAttribute("Mask", …)` harder.** Zero code. Leaves a silent discard in place, which this repo treats as a bug rather than a documentation gap (#184, #191, #192, #211 were all filed against precisely that).

### Recommendation

**A.** B leaves the silent-discard half of the problem live, and a second issue to close it later would touch the same three files again. The `IMask` overload is the smaller half of A once the constants exist: `Resolve` prefers a supplied instance and otherwise builds a `PatternMask` from the pattern.

On instance sharing, the honest answer is that a supplied `IMask` is safe *as a carrier* — `MudMask.SetMask` copies out of it via `UpdateFrom` when the types match and never retains it — but only while the caller does not also mutate it. Document it as a prototype, and note the type-stability rule #211 established: the resolved type must not vary per render, or `SetMask` takes its adopt-outright branch mid-edit.

</details>

<details>

### Goal

A discoverable, type-safe way to configure a mask in `FormCraft.ForMudBlazor` that can express delimiter stripping and any MudBlazor `IMask`, without changing the core contract.

### Scope

- `.WithMask(string pattern)` and `.WithMask(string pattern, bool cleanDelimiters)` on `MudBlazorFieldBuilderExtensions`.
- `.WithMask(IMask mask)` for a pre-built MudBlazor mask.
- `TextMaskMap.Resolve` reads the new attributes; both render paths keep calling it unchanged.
- Named constants for every attribute key, following `AdornmentClickAttribute`.

### Non-goals

- Changing `ITextFieldComponent.Mask` (stays `string?`, stays framework-agnostic).
- Removing or deprecating `.WithAttribute("Mask", "…")` — additive only.
- A diagnostic for a value the mask blanks — separate follow-up.
- Numeric/date masks. This is the text path only.

### Behaviour

| Configuration | Resolved mask | Model receives |
|---|---|---|
| nothing | `null` | value unchanged |
| `.WithMask("0000-0000")` | `PatternMask("0000-0000")` | `"1234-5678"` (delimited) |
| `.WithMask("0000-0000", cleanDelimiters: true)` | `PatternMask` with `CleanDelimiters = true` | `"12345678"` |
| `.WithMask(new RegexMask(…))` | that instance | per that mask |
| `.WithAttribute("Mask", "0000-0000")` | `PatternMask("0000-0000")` | `"1234-5678"` — unchanged |
| blank/whitespace pattern | `null` | value unchanged (#211 rule) |

Identical on both render paths — that is what `RenderPipelineParityTests` exists to enforce.

### Key files

- `FormCraft.ForMudBlazor/Extensions/FieldBuilderExtensions.cs` — the new overloads + constants.
- `FormCraft.ForMudBlazor/Fields/TextField/MudBlazorTextFieldComponent.razor.cs` — `TextMaskMap.Resolve`.
- `FormCraft.ForMudBlazor/Features/CollectionField/CollectionFieldComponent.razor.cs` — `AddTextInputAttributes`.
- Tests: `MudBlazorTextFieldComponentTests.cs`, `CollectionInputTypeTests.cs`, `RenderPipelineParityTests.cs`.
- `README.md` — the `## 🎉 Unreleased` section.

### Validation rules

- A blank or whitespace-only pattern resolves to `null`, per #211 — an empty `PatternMask("")` would reroute an unmasked field through `MudMask` and drop `MaxLines` with it.
- A supplied `IMask` wins over a pattern string if both are somehow set; a pattern string wins over nothing.
- `cleanDelimiters` is meaningless without a pattern and is ignored there.

### Edge cases

- **Type stability.** `Resolve` must return the same `IMask` implementation on every render for a given field. `MudMask.SetMask` preserves the user's text and caret only via `_mask.UpdateFrom(other)` when types match, and adopts the new instance outright otherwise — so varying the type mid-edit resets the field. Already pinned by `Mask_Should_Resolve_Identically_On_Both_Paths`.
- **A supplied `IMask` is a prototype, not the live mask.** MudBlazor copies settings out of it; it is never the object the input edits. Safe to share across rows, but callers must not mutate it after configuring.
- **`cleanDelimiters: true` changes what the model stores**, so it is opt-in and the default keeps #211's behaviour.

### Assumptions

- `.WithAttribute("Mask", …)` stays supported indefinitely; this is additive, not a migration.
- Only the text path gains this. Numeric and date fields do not consult `TextMaskMap`.
- MudBlazor 9.8.0 semantics (`PatternMask`, `CleanDelimiters`, `SetMask`) as measured in #253.

</details>

## 🛠️ Implementation plan

**Goal:** a typed `.WithMask(...)` builder covering pattern, `cleanDelimiters`, and a pre-built `IMask`.
**Architecture:** fluent builder extension in the MudBlazor adapter → attribute constants → `TextMaskMap.Resolve` → both render paths.

**Global Constraints**

- **Keep `FormCraft` (core) UI-agnostic** — no MudBlazor type may appear in it. `ITextFieldComponent.Mask` stays `string?`.
- **Both render paths move together.** Anything `TextMaskMap.Resolve` learns must be reachable from `MudBlazorTextFieldComponent` *and* `CollectionFieldComponent.AddTextInputAttributes` in the same change (#203 is the standing root issue for why).
- **`TextMaskMap.Resolve` must return a stable `IMask` type per field across renders** — see Edge cases.
- **Sequence numbers** in `CollectionFieldComponent` are a shared index space; the text block starts at `TextAttributeStart` and `SequenceNumberBudgetTests` enforces it.
- **`dotnet test --filter` is inert here** (Microsoft.Testing.Platform prints `MTP0001` and runs everything) — run `dotnet test -c Release` and read the summary.
- **Build in Release**: `Directory.Build.props` sets `TreatWarningsAsErrors=true`; do not relax it.

### Task 1: `.WithMask(pattern)` and `.WithMask(pattern, cleanDelimiters)` on the component path

**Files:** modify `FormCraft.ForMudBlazor/Extensions/FieldBuilderExtensions.cs`, `FormCraft.ForMudBlazor/Fields/TextField/MudBlazorTextFieldComponent.razor.cs`; test `FormCraft.ForMudBlazor.UnitTests/Fields/MudBlazorTextFieldComponentTests.cs`.

**Interfaces:** `WithMask<TModel>(this FieldBuilder<TModel, string> field, string pattern, bool cleanDelimiters = false)`; `TextMaskMap.CleanDelimitersAttribute`; `TextMaskMap.Resolve` gains the options.

- [x] **Step 1:** Write failing tests: `.WithMask("0000-0000")` binds a `PatternMask` with that pattern and `CleanDelimiters == false`; `.WithMask("0000-0000", cleanDelimiters: true)` binds one with `CleanDelimiters == true`.
- [x] **Step 2:** Run `dotnet test -c Release` → FAIL (method not found).
- [x] **Step 3:** Add the `WithMask` overloads next to `WithNativeRequired`, writing `TextMaskMap.AttributeName` and a new `CleanDelimitersAttribute` constant.
- [x] **Step 4:** Extend `TextMaskMap.Resolve` to set `CleanDelimiters` on the `PatternMask` it builds, keeping the blank-pattern→`null` rule.
- [x] **Step 5:** Read the flag on the component path and pass it to `Resolve`.
- [x] **Step 6:** Run `dotnet test -c Release` → PASS.
- [x] **Step 7:** Commit: `feat(mudblazor): add a typed WithMask builder with delimiter control`

### Task 2: Honour the new attributes on the collection item path

**Files:** modify `FormCraft.ForMudBlazor/Features/CollectionField/CollectionFieldComponent.razor.cs`; test `FormCraft.ForMudBlazor.UnitTests/Fields/CollectionInputTypeTests.cs`, `FormCraft.ForMudBlazor.UnitTests/Components/RenderPipelineParityTests.cs`.

**Interfaces:** `AddTextInputAttributes` reads the same two constants and calls the same `Resolve`.

- [x] **Step 1:** Write a failing parity test: `.WithMask(pattern, cleanDelimiters: true)` resolves to the same pattern **and** the same `CleanDelimiters` on both paths.
- [x] **Step 2:** Run `dotnet test -c Release` → FAIL (item path ignores the flag).
- [x] **Step 3:** Read the options attribute in `AddTextInputAttributes` and pass them to `Resolve`, without adding an `AddAttribute` call (reuse the existing `Mask` slot).
- [x] **Step 4:** Run `dotnet test -c Release` → PASS, `SequenceNumberBudgetTests` included.
- [x] **Step 5:** Commit: `fix(mudblazor): honour mask options on collection item fields`

### Task 3: `.WithMask(IMask)` for pre-built MudBlazor masks

**Files:** modify `FormCraft.ForMudBlazor/Extensions/FieldBuilderExtensions.cs`, `MudBlazorTextFieldComponent.razor.cs`, `CollectionFieldComponent.razor.cs`; test all three MudBlazor test files above.

**Interfaces:** `WithMask<TModel>(this FieldBuilder<TModel, string> field, IMask mask)`; `TextMaskMap.MaskInstanceAttribute`.

- [x] **Step 1:** Write failing tests on both paths: `.WithMask(new RegexMask(@"^\d{4}$"))` binds that mask (assert the type and pattern, not the instance).
- [x] **Step 2:** Run `dotnet test -c Release` → FAIL (silently discarded today).
- [x] **Step 3:** Add the overload writing `MaskInstanceAttribute`, and make `Resolve` prefer a supplied instance over a pattern string.
- [x] **Step 4:** Document on the overload that the argument is a **prototype** MudBlazor copies settings from, and must not be mutated after configuration.
- [x] **Step 6:** Commit: `feat(mudblazor): accept a pre-built IMask in WithMask`

### Task 4: Document the builder and the delimiter behaviour

**Files:** modify `README.md`; XML docs on the new overloads.

- [x] **Step 1:** Add a `## 🎉 Unreleased` entry: the typed builder, `cleanDelimiters` as the answer to "the model stores the masked text" from #211, and the `IMask` overload.
- [x] **Step 2:** Update the #211 entry's point 1, which currently says `CleanDelimiters` cannot be reached.
- [x] **Step 3:** Run `dotnet build -c Release` and `dotnet test -c Release` → both green.
- [x] **Step 4:** Commit: `docs: document the typed WithMask builder`

phmatray:
Implements #265.

Closes #265.

Replaces the `.WithAttribute("Mask", "…")` magic string with a discoverable, type-safe
`.WithMask(...)` builder in the MudBlazor adapter — reaching `PatternMask.CleanDelimiters`
and pre-built `IMask` instances, both unreachable today. Core stays UI-agnostic:
`ITextFieldComponent.Mask` remains `string?`.

code-review pass.

### Plan
- [x] Task 1: `.WithMask(pattern)` and `.WithMask(pattern, cleanDelimiters)` on the component path
- [x] Task 2: Honour the new attributes on the collection item path
- [x] Task 3: `.WithMask(IMask)` for pre-built MudBlazor masks
- [x] Task 4: Document the builder and the delimiter behaviour

### Plan deviations

The plan was written before #250 landed on `dev`; two of its premises no longer hold, so
Task 2 shipped differently from what it prescribes.

1. **`CollectionFieldComponent.AddTextInputAttributes` no longer exists.** #203/#250 deleted the
   hand-rolled `RenderTreeBuilder` path — item fields now render through `IFieldRendererService`,
   i.e. the *same* `MudBlazorTextFieldComponent` a standalone field uses, reading the same
   `AdditionalAttributes`. Task 1's attribute is therefore already honoured on the item path, and
   Task 2 had no second implementation to correct. Its expected red bar (Step 2) never
   materialised: the parity test passed the moment it was written.
2. **`SequenceNumberBudgetTests` (Task 2, Step 4) does not exist on `dev`.** It lives on the
   unmerged `pr254` branch, so it could not be included in the run.

Task 2 therefore ships as the parity assertion alone, committed as `test:` rather than the plan's
`fix:` — nothing was repaired, and on a squash-merging repo a `fix:` subject would put a false
claim into the history release-please parses. The test is not vacuous: reverting Task 1's flag
forwarding turns its `cleanDelimiters: true` case red, verified by mutation before committing.

3. **⚠️ `.WithMask` takes `Func<IMask>`, not `IMask` — please sanity-check this call.** The issue
   specifies `.WithMask(new RegexMask(…))`, and its brainstorm reasoned that a supplied instance is
   "safe *as a carrier* — `MudMask.SetMask` copies out of it via `UpdateFrom` … and never retains
   it". **That premise is wrong**, and the code review caught it. `SetMask` only copies when the
   incoming mask matches the type it already holds; `MudMask` seeds itself with a `PatternMask`, so
   a `RegexMask`/`BlockMask`/`MultiMask` — precisely the types this overload exists to reach — hits
   `other.SetText(ReadText); _mask = other;` and MudBlazor **keeps and mutates the caller's
   object**.

   That matters because one `IFieldConfiguration` is shared by every row of a collection. I
   confirmed it empirically before changing anything: with the instance overload, two rows received
   the identical `IMask`, and it was the caller's own object. Since `BaseMask` carries the live
   `Text`/`CaretPos`/`Selection` of one input, that is two rows sharing one editing state — and it
   breaks the fresh-instance-per-call invariant `TextMaskMap` documents at length for the pattern
   path.

   `Func<IMask>` is invoked per rendered field, which fixes it and keeps the built configuration
   immutable as `CLAUDE.md` requires. Cloning an arbitrary `IMask` instead was rejected:
   `RegexMask`'s parameterless constructor is `protected`, so there is no general copy.

   The cost is ergonomic — `() => new RegexMask(…)` rather than `new RegexMask(…)` — and it is a
   deliberate departure from the issue's stated signature, so it is the one decision here worth a
   second opinion before merge.

### Code review

Run at `high` over the whole branch; 11 findings, all triaged. Fixed in `3665b36`: the shared-mask
defect above; the blank-pattern rule now also covers a factory-produced mask; the two overloads
clear each other's keys so the last `WithMask` wins in **both** orders (previously
instance-always-wins, which would silently ignore a caller's later override); `Resolve`'s new
parameters lost their defaults so a future reader cannot drop them silently; the two shuttle
properties are `private` rather than new public API surface; `TextMaskMap`'s summary no longer
describes the render path #203/#250 deleted; and the README's "prototype … MudBlazor does not hold
on to it" claim — false for those mask types — is replaced by the factory rationale.

Two tests the review identified as missing were added: the end-to-end assertion that
`cleanDelimiters: true` really writes `5551234567` to the model (the README's headline promise,
previously only asserted on the bound parameter), and a two-row collection test that fails without
the factory fix.

Not fixed, by judgement: mixing the typed builder with a *later* raw `.WithAttribute("Mask", …)` on
the same field leaves the earlier `cleanDelimiters` in effect, since the raw escape hatch cannot
clear a companion key. The README now states this rather than claiming the raw form is unconditionally
unchanged.


phmatray:
Recording one behavioural detail from the review of this PR, deliberately **not** filed as an issue — it is documented in the README and only reachable by mixing the two configuration surfaces on a single field.

`.WithMask(pattern, cleanDelimiters)` and `.WithMask(Func<IMask>)` clear each other's attribute keys, so the last call wins in both directions. The legacy `.WithAttribute("Mask", "…")` escape hatch cannot participate in that: it sets only the pattern key, so a `cleanDelimiters: true` from an earlier `.WithMask(...)` call on the same field stays in effect.

Using the raw form on its own is unaffected (`CleanDelimiters` defaults to `false`). Prefer `.WithMask(...)`, which overrides whatever the other overload configured.

The mask-blanks-a-stored-value diagnostic that #265's spec deferred is tracked separately in #266.

Issue #256, opened 2026-08-12
Test reports are named by runner and timestamp, so a downloaded artifact cannot be attributed to a suite


phmatray:
## Problem / motivation

Since #231 (PR #252) the `Test` target drives Microsoft.Testing.Platform's own
`--results-directory` / `--report-xunit-trx` / `--report-xunit-html`, so `test-results/` is genuinely
populated and all three workflows upload something real. What lands there is not equally readable.

The solution has **two** test assemblies writing into **one** directory, and the trx/html reports take
the runner's default name — `<user>_<machine>_<timestamp>`. Measured on the real CI artifact from
PR #252's `build-and-test` run:

```
FormCraft.ForMudBlazor.UnitTests_net10.0_x64.log     <- named
FormCraft.UnitTests_net10.0_x64.log                  <- named
runner_runnervmvrwv9_2026-08-12_12_42_45.948242.trx  <- which suite?
runner_runnervmvrwv9_2026-08-12_12_42_45.948242.html
runner_runnervmvrwv9_2026-08-12_12_42_45.995919.trx  <- which suite?
runner_runnervmvrwv9_2026-08-12_12_42_45.995919.html
```

The two trx files differ **only in the microsecond field**. The per-assembly `.log` files in the same
artifact are named properly, which makes the inconsistency sharper: half the artifact tells you where
it came from and half does not.

The practical cost is small but real, and lands at the worst moment. Someone downloads the artifact
because CI went red and they want to know which suite broke; the `.trx` is the machine-readable record
of exactly that, and they have to open both to find out which is which. Attribution is *recoverable* —
each trx records fully-qualified test names, so the assembly is inferable from the contents — but not
from the file name, which is what a human scans first.

### Why the obvious fix is wrong

`--report-xunit-trx-filename` sets a **fixed** name. With one `dotnet test` invocation over the
solution and both assemblies pointed at one results directory, the second assembly would overwrite the
first — silently losing a whole suite's report, which is strictly worse than an opaque name. That is
why #252 deliberately kept the default naming, and the reasoning is recorded in `build/Build.cs`.

So the fix has to change the *directory*, not the filename.

## Proposed solution

Give each test project its own results subdirectory — `test-results/<ProjectName>/` — by invoking
`DotNetTest` once per test project instead of once for the solution. Uniqueness is preserved (separate
directories), and identity comes from the path.

Knock-on effects to handle in the same change:

[...]

Issue #255, opened 2026-08-12
The CI guard tests hand-parse workflow YAML, and now do it twice


phmatray:
## Problem / motivation

`FormCraft.UnitTests/Ci/` now holds two guard suites that read the same files the same way, with their
own copies of the code that does it.

`TrustedPublishingWorkflowTests.cs` (#173, #198, #226) and `TestReportingTests.cs` (added in #252) each
declare their own:

- `RepoRoot` / `LocateRepoRoot()` — walk up from `AppContext.BaseDirectory` until `FormCraft.sln`
- `WorkflowsDirectory`, `ReadWorkflow()`, `ReadBuildScript()`
- `WithoutComments(text, marker)` — drop whole-line comments so a "not referenced" assertion fires on
  wiring rather than on prose
- an independent enumeration of `.yml`/`.yaml` under `.github/workflows`

These are byte-identical **including their explanatory comments** — both carry the same
`dir.Parent is DirectoryInfo?, so the variable has to be too` note and the same
`a trailing-comment strip would also mangle the "https://" in a URL` justification. And
`TestReportingTests.StepNamed()` is a one-line variant of `TrustedPublishingWorkflowTests.StepWithId()`:
both find a step by scanning from its header line to the next line whose trimmed text starts with
`- `, differing only in whether the header is matched on `- name:` or on `id:`.

This is not a tidiness complaint. That step-boundary scan is an **approximation of YAML**, shared by
both files and owned by neither. A fix to it now has to be made twice or the copies drift — and this
codebase has already paid that price: #205 records a copied test comment that propagated a factual
error (`AddCommonFieldAttributes` described as "shared by the text and numeric paths") and hid the date
path from coverage until review caught it. **Duplicated scaffolding duplicates its errors too.**

### The gap the duplication is holding open

Both suites' assertions are **file-scoped, not job-scoped**, and closing that is the same piece of work.

`WorkflowsThatRunTests()` matches on file *content*, and `StepNamed()` scans the *whole file*. So a
workflow that ran `./build.cmd Test` in job A and uploaded the `test-results` artifact from job B would
satisfy every assertion in `TestReportingTests` while uploading nothing at all — job B gets a fresh
workspace, so `test-results/` never exists there. Combined with `if-no-files-found: ignore` (which
#252 pinned on all three uploads), that failure is completely silent.

This is not hypothetical: `release-please.yml` is already a two-job workflow (`release-please` and
[...]

Issue #226, opened 2026-08-11
Decide whether a missing NUGET_USER should silently skip the release publish


phmatray:
## Problem / motivation

The NuGet login step in `.github/workflows/release-please.yml` is gated on **two** conditions:

```yaml
      - name: NuGet login (OIDC -> short-lived key)
        id: login
        if: ${{ needs.release-please.outputs.release_created == 'true' && env.NUGET_USER != '' }}
```

The first clause is right and load-bearing. **The second clause is the shape #198 explicitly warned
against**, quoted from that issue's brainstorm:

> The runbook's rule is that a tag-only workflow must *fail loudly* (no `if: env.NUGET_USER != ''`
> guard), because a guard on a tag produces a green CI with an unpublished version that nobody
> notices. […] **Do not "improve" this into a `NUGET_USER` guard.**

Concretely, if `NUGET_USER` is ever unset, renamed, or scoped away:

1. the login step **skips** rather than fails;
2. `steps.login.outputs.NUGET_API_KEY` is the empty string, so Nuke's `Publish` never runs;
3. release-please still cuts the version, the tag, the changelog and the GitHub Release;
4. **the workflow is green.**

The result is a released version that exists on GitHub and in the changelog but **not on nuget.org**,
with nothing anywhere reporting a problem. That is the worst failure shape for a release path that
runs a handful of times a year — the same "green run, unpublished version" the parent issue exists to
prevent.

This is **not** presented as a defect to fix blindly. #202's context is genuinely narrower than the
case the warning came from: the clause is `&&`-ed with `release_created`, and it lives in a
release-only workflow rather than the mixed push/PR/tag workflow the rule was written about. So this
is a judgement call the owner should make deliberately, rather than something that should sit
unexamined because it was introduced in passing.

Note that #200's guard test deliberately asserts on `release_created` and **not** on the
`NUGET_USER` clause — so removing the clause will not turn the suite red. That was intentional, to
leave this decision open.

## Proposed solution

Decide explicitly between two coherent designs, then encode the decision so it stops being ambiguous:

- **Fail loudly (recommended).** Drop `&& env.NUGET_USER != ''`. A missing secret then fails the
  login step, which fails the job, which makes the missed publish impossible to overlook.
- **Keep the guard, but make the silence impossible.** Retain the clause and add a step that fails
  the job when a release was created but no package was pushed, so "released but unpublished" is
  still loud.

Either way, record the reasoning in a comment on the step, so the next reader does not re-litigate it
[...]

Issue #198, opened 2026-08-11
Prove NuGet Trusted Publishing end-to-end and retire the long-lived NUGET_API_KEY


phmatray:
## Problem / motivation

FormCraft's release path was rewired to publish via **NuGet Trusted Publishing (OIDC)** in #173
(merged 2026-07-27). The workflow side is done and on `dev` today: `.github/workflows/continuous.yml`
declares `id-token: write`, runs `NuGet/login` pinned by digest (`8d19675…`, v1.2.0) gated on
`startsWith(github.ref, 'refs/tags/v')`, and feeds the short-lived key to Nuke as `NUGET_API_KEY`.
The `NUGET_USER` secret is set, and the nuget.org policy was registered the same day.

**But that path has never once executed, and the long-lived key it was meant to retire is still there.**

Measured on 2026-08-11:

| Fact | Value | How it was measured |
|---|---|---|
| Last version tag | `v3.1.0`, **2026-06-10** | `git log -1 --format=%ci v3.1.0` |
| Trusted Publishing merged | **2026-07-27T18:48Z** (#173) | `gh pr view 173 --json mergedAt` |
| Tag-triggered workflow runs since | **none** | `gh run list` filtered on `event=push`, `headBranch` starting `v` |
| Latest version on nuget.org | `3.1.0` (both packages) | `api.nuget.org/v3-flatcontainer/…/index.json` |
| `NUGET_API_KEY` repo secret | **still present**, created 2026-02-23 | `gh secret list` |
| `NUGET_USER` repo secret | present, created 2026-07-27 | `gh secret list` |
| Org-scope secrets | **n/a** — `phmatray` is a user account (API returns 422) | `gh api …/actions/organization-secrets` |

So the last tag predates the rewrite by six weeks. Every published version of `FormCraft` and
`FormCraft.ForMudBlazor` went out under the old long-lived key, and the OIDC path is **wired but
unproven**. Two consequences, both bad:

1. **The next release is the test.** If the policy fields, the workflow filename, or the Nuke
   `Publish` chain are wrong, we find out at the exact moment we want to ship a ★61 flagship
   package — and a half-published release on nuget.org cannot be undone (nuget.org delists, it
   never deletes).
2. **The credential the migration was supposed to eliminate is still live.** Nothing references
   `secrets.NUGET_API_KEY` in the workflow any more, so it is *inert*, not *armed* — but an inert
   long-lived push credential is still a long-lived push credential, and deleting the GitHub secret
[...]

PR #287, opened 2026-08-13
test(ci): make the shared step scan bind to exactly one step (#267)

### LINKED ISSUE #267: Make the shared step-extraction primitives precise, and close the YAML-parser question
## Problem / motivation

#255 consolidated the `Ci/` guard suites onto one reader, `FormCraft.UnitTests/Ci/WorkflowSource.cs`.
Its two step-extraction primitives now sit side by side in that one file — and they do not agree with
each other about how precise a step match has to be.

**`StepWithId` matches `id: <x>` as an unanchored substring, anywhere in a line:**

```csharp
var start = Array.FindIndex(lines, l => l.Contains($"id: {stepId}", StringComparison.Ordinal));
```

So a search for `login` also matches `id: login-legacy`, `id: login2`, and any `with:` value or comment
line that happens to contain the text `id: login`. `Array.FindIndex` takes the **first** such line, so
an unrelated earlier line silently redefines the whole slice — and every assertion about the login
step's `if:` and `uses:` is then answered by the wrong step. That is the failure mode
`StepWithId` exists to prevent, reintroduced one level down.

Its sibling `StepNamed` is prefix-anchored (`TrimStart().StartsWith($"- name: '{stepName}'")`). Two
primitives, one shared reader, two different standards of precision — and the *looser* one guards the
release path (#226), where a wrong answer is a green run on a version that never reached nuget.org.

**The slice also starts at the wrong line.** `StepWithId` scans from the `id:` line rather than from
the step's own `- name:` header, so the returned "step" excludes the step's name — it cannot be used
to assert anything *about* the name, and it silently differs in shape from what `StepNamed` returns
for the same step.

**And a missing step reports itself five times.** `StepNamed` asserts internally
(`start.ShouldBeGreaterThanOrEqualTo(0, …)`) rather than returning `null`. Delete the
`Publish: test-results` step from a job and `Every_Job_That_Runs_Tests_Should_Upload_The_TestResults_Artifact`
reports the offender cleanly — while the other four tests each throw from inside the shared helper,
producing four more red tests that describe the same single root cause as an apparent helper failure.

### Why now

#255's *Alternatives considered* parked the YAML-parser option with an explicit precondition:

> **Add a real YAML parser** (YamlDotNet) … Rejected for now … B stays available and becomes an easier
> call once the reader is a single type with a single set of callers — which is itself an argument for
> doing A first.

That precondition is now met. So this issue is not only "fix two bugs" — it is the point at which the
parked decision gets made deliberately, rather than by drift, with the concrete defects above as the
evidence for what the hand-rolled scan actually costs.

## Proposed solution

Keep the hand-rolled scan and make it precise, and close the parser question explicitly rather than
leaving it open a third time.

1. Anchor `StepWithId` on an exact `id: <stepId>` key match instead of a substring, and make an
   ambiguous match (two steps claiming one id) **fail loudly** rather than resolve to the first.
2. Start the returned slice at the step's own list-item header, so `StepWithId` and `StepNamed` return
   the same shape for the same step.
3. Add `TryStepNamed`/`TryStepWithId` returning `string?`, and route the five upload assertions through
   them so a missing step is reported once, uniformly, by each test that cares.

The parser stays rejected, and this issue records *why* so it stops being re-litigated: several
assertions deliberately work on raw text — `WithoutComments` exists precisely so a documentation-only
edit cannot redden the suite, and claims like `uses:\s*NuGet/login@[0-9a-f]{40}\b` are textual by
design. A parsed model discards exactly that, so adopting one means carrying **both** models, not
replacing one with the other.

## Alternatives considered

- **Adopt YamlDotNet and parse properly.** Correct by construction, and step/job location falls out for
  free. Rejected: it adds a centrally-pinned dependency (`Directory.Packages.props`) to assert on five
  files the repo controls, and — the deciding factor — it does not *remove* the raw-text model, because
  `WithoutComments` and the "is this option enabled" assertions must keep working on text. Two models
  in one reader is worse than one imprecise model made precise. Reopen this if the assertions ever turn
  materially structural (needing `needs:` graphs, matrix expansion, or anchor/alias resolution) — that
  is the criterion, not general tidiness.
- **Fix only the `id:` anchoring, leave the reporting alone.** Smallest diff, and the anchoring bug is
  the one with real consequences. Rejected as a stopping point: the four-extra-failures behaviour is
  what makes a genuine breakage hard to read, and both live in the same handful of lines now.
- **Leave it.** Both weaknesses are pre-existing and both suites are green. But #255's whole premise is
  that an approximation owned by nobody drifts; it now has an owner, and this is the first change since
  it acquired one.

## Area

Test infrastructure — `FormCraft.UnitTests/Ci/WorkflowSource.cs` and its two callers. No library code,
no workflow, no `build/Build.cs` behaviour change.

**Related:** #255 (created the shared reader; this is its deferred tail), #257 (the PR that landed it
and surfaced these), #226 (the assertion the `id:` anchoring protects), #198 (owns the sibling
job-scoping follow-up), #205 (the same duplicated-scaffolding pattern in the MudBlazor test project).

<details>

### Problem / context

`FormCraft.UnitTests/Ci/` asserts on repository *files* rather than library behaviour, because the
things it guards run a handful of times a year (a release, a publish) or fail silently (#231's inert
reporting). #255 removed the duplication across those suites by extracting `WorkflowSource`.

Extraction moved the primitives unchanged — deliberately, since #255's plan forbade changing what
either suite asserts. That was the right call for that change and it is what leaves this one to do:
the two step scans are now visibly adjacent, and visibly inconsistent.

The consequence is not theoretical. `StepWithId` is the primitive #226 depends on, and #226 exists
because gating the NuGet login on the wrong thing produces a **green run on a version that never
reached nuget.org**. A primitive that can bind to the wrong step is a bad foundation for that claim.

### Approaches

**A. Harden the hand-rolled scan.** Anchor the `id:` match, include the step header, add `Try*`
variants.
*Trade-offs:* small, no dependency, and every change is local to one file with two callers. Leaves the
scan an approximation of YAML — but a materially tighter one, and the repo's five workflow files are
its only inputs.

**B. Replace the scan with YamlDotNet.** Parse into a model; locate jobs and steps structurally.
*Trade-offs:* correct by construction for structure. But the raw-text assertions do not go away, so
the reader ends up holding a parsed model *and* the text one — and it spends a centrally-pinned
dependency to do it. #255 rejected this on the same grounds; nothing has changed except that the
consolidation precondition is now met, which lowers the *cost* of adopting it without changing the
*reason* against.

**C. Parse for location, keep text for assertions (explicit hybrid).** Make B's dual model a stated
design rather than an accident.
*Trade-offs:* honest about what B actually is, and genuinely the right end state if these assertions
keep growing structurally. Still pays the dependency, for a subset of A's benefit today.

### Recommendation

**A**, with the parser question closed in writing and a named criterion for reopening it (assertions
becoming materially structural — `needs:` graphs, matrix expansion, anchors). The defects are a
handful of lines in one file; the dependency is forever, and would not retire the text model it is
supposed to replace.

**Assumptions:** both suites stay green throughout — this tightens *how* a step is located, not *what*
is asserted about it, so a red bar means the tightening bound to a different step than before and is a
finding, not a formatting problem. No new package. `release-please.yml` remains the fixture, being the
only workflow with `id:`-bearing steps.

</details>

<details>

### Goal

Step extraction that cannot bind to the wrong step, returns a consistent shape from both primitives,
and reports a missing step once per test that cares.

### Scope

- `FormCraft.UnitTests/Ci/WorkflowSource.cs` — anchoring, slice start, `Try*` variants.
- `FormCraft.UnitTests/Ci/WorkflowSourceTests.cs` — pin each new edge.
- `FormCraft.UnitTests/Ci/TestReportingTests.cs` — route the five upload assertions through `Try*`.

### Non-goals

- No new NuGet dependency; no YAML parser.
- No change to any workflow, to `build/Build.cs`, or to library code.
- No change to *what* either suite asserts — only to how a step is located and how absence is reported.
- Job-scoping `TrustedPublishingWorkflowTests` is **not** here; it is tracked on #198.

### Surface

```
internal static class WorkflowSource
    string? TryStepNamed (string scope, string stepName)        // null when absent
    string? TryStepWithId(string scope, string stepId)          // null when absent
    string  StepNamed    (string scope, string stepName, string? scopeDescription = null)   // asserts, as today
    string  StepWithId   (string scope, string stepId,   string? scopeDescription = null)   // asserts, as today
```

The asserting forms stay, implemented over the `Try*` forms, so existing call sites are unchanged.

### Matching rules

```mermaid
flowchart TD
    A[scope text] --> B[find lines whose trimmed text is exactly 'id: &lt;stepId&gt;']
    B --> C{how many?}
    C -->|0| D[return null]
    C -->|2+| E[fail loudly: ambiguous id]
    C -->|1| F[walk BACK to nearest line whose trimmed text starts with '- ']
    F --> G[slice from that header to the next '- ' line]
```

- **Exact key match**, not `Contains`: the trimmed line must be `id: <stepId>`, allowing trailing
  whitespace and a trailing `# comment`. This is what stops `login` matching `login-legacy`.
- **Ambiguity is loud.** Two steps claiming one id is invalid in GitHub Actions; resolving it to the
  first is how the current code returns the wrong step silently.
- **Walk back to the step header** so the slice begins at the step's own `- ` list item — its `- name:`
  where it has one, its `- uses:` where it does not. Both primitives then return the same shape.

### Edge cases

- A step whose `id:` precedes its `- name:` — impossible, `- ` starts the item; the walk-back handles
  any key order within the step.
- An `id:`-bearing step that is the last in its scope → slice runs to the end (already handled).
- `StepCondition` in `TrustedPublishingWorkflowTests` picks the first line starting with `if:`; the
  added header line is a `- name:`/`- uses:` line, so it is unaffected.
- A scope with no `- ` line before the match (malformed) → treat as absent rather than slicing from 0.

### Assumptions

Including the header line can only *add* text to a slice that assertions currently `ShouldContain`, so
no existing positive assertion can break. The one to watch is
`TrustedPublishingWorkflowTests.ReleaseWorkflow_Should_Not_Gate_The_Login_On_The_NuGetUser_Secret`,
which asserts a `ShouldNotContain` — on the `if:` line only, via `StepCondition`, so it is unaffected.

</details>

## 🛠️ Implementation plan

**Goal:** step extraction that cannot bind to the wrong step, with uniform absence reporting.

repository files (`.github/workflows/*.yml`, `build/Build.cs`) rather than on library behaviour.

**Global Constraints**

- `TreatWarningsAsErrors=true` is deliberate — never relax it to make a build pass. Nullability
  matters here: the `Try*` members return `string?` and the asserting wrappers must not leak that.
- **No new NuGet package.** Adding a YAML parser is explicitly out of scope (see *Alternatives*).
- No change to any workflow, to `build/Build.cs`, or to library code. If an assertion would require
  editing a workflow to pass, stop — that is a finding, not a refactor.

### Task 1: Anchor `StepWithId` on an exact `id:` match, and make ambiguity loud

**Files:** modify `FormCraft.UnitTests/Ci/WorkflowSource.cs`,
`FormCraft.UnitTests/Ci/WorkflowSourceTests.cs`.

**Interfaces:** no public change yet — `StepWithId(scope, stepId, scopeDescription)` keeps its
signature; only how it locates the line changes.

- [x] **Step 1:** Write the failing tests in `WorkflowSourceTests.cs`: against a small inline fixture
      containing both `id: login-legacy` (first) and `id: login` (second), `StepWithId(fixture, "login")`
      must return the **`login`** step, not the legacy one; and a fixture with two `id: login` steps must
      fail with an "ambiguous" message. Run `dotnet test -c Release` → **RED**.
- [x] **Step 2:** Replace `l.Contains($"id: {stepId}")` with an exact trimmed-key match tolerating
      trailing whitespace and a trailing `# comment`, and assert exactly one match before slicing.
- [x] **Step 3:** Run `dotnet test -c Release` → **GREEN**, with `TrustedPublishingWorkflowTests`
      unchanged (its `login` step is still found).
- [x] **Step 4:** Commit: `test(ci): anchor the step-id scan so it cannot bind to a longer id`

### Task 2: Start the slice at the step's own list-item header

**Files:** modify `FormCraft.UnitTests/Ci/WorkflowSource.cs`,
`FormCraft.UnitTests/Ci/WorkflowSourceTests.cs`.

**Interfaces:** unchanged; `StepWithId` now returns the same shape as `StepNamed` for the same step.

- [x] **Step 1:** Write the failing test: `StepWithId(Read("release-please.yml"), "login")` must contain
      `- name: NuGet login`, and must still exclude the following step's `- name: 'Run: Pack`. Run
      `dotnet test -c Release` → **RED**.
- [x] **Step 2:** Walk back from the matched `id:` line to the nearest line whose trimmed text starts
      with `- `, and slice from there. Treat "no header found" as absent rather than slicing from 0.
- [x] **Step 3:** Run `dotnet test -c Release` → **GREEN**. Confirm
      `ReleaseWorkflow_Should_Not_Gate_The_Login_On_The_NuGetUser_Secret` still passes — it reads only
      the `if:` line via `StepCondition`, so the added header must not reach it.
- [x] **Step 4:** Commit: `test(ci): return a step from its own header so both scans agree in shape`

### Task 3: Add `Try*` variants and report a missing step uniformly

**Files:** modify `FormCraft.UnitTests/Ci/WorkflowSource.cs`,
`FormCraft.UnitTests/Ci/WorkflowSourceTests.cs`, `FormCraft.UnitTests/Ci/TestReportingTests.cs`.

**Interfaces:** add `string? TryStepNamed(string scope, string stepName)` and
`string? TryStepWithId(string scope, string stepId)`; reimplement the asserting forms over them.

- [x] **Step 1:** Write the failing test: `TryStepNamed(scope, "nope")` returns `null` rather than
      throwing, while `StepNamed(scope, "nope")` still fails with a message naming the scope. Run
      `dotnet test -c Release` → **RED**.
- [x] **Step 2:** Add both `Try*` members and reimplement `StepNamed`/`StepWithId` over them, keeping
      their assertion messages byte-identical so existing failures read the same.
- [x] **Step 3:** Run `dotnet test -c Release` → **GREEN**.
- [x] **Step 4:** Route `TestReportingTests`' five upload assertions through `TryStepNamed`, so a job
      missing the step is reported as an offender by each test rather than throwing from the helper.
      Keep the `(workflow, job)` offender labelling and the vacuity guard.
- [x] **Step 5:** Run `dotnet test -c Release` → **GREEN** with an unchanged pass count.
- [x] **Step 6:** Prove the reporting actually improved: temporarily delete the
      `Publish: test-results` step from `ci.yml`, confirm all five tests now name
      `ci.yml / build-and-test` as an offender rather than four of them throwing, then revert.
- [x] **Step 7:** Commit: `test(ci): report a missing upload step uniformly across the guards`

phmatray:
Implements #267.

Closes #267.

Makes the two step-extraction primitives on the shared `Ci/` reader (`WorkflowSource`, #255) agree
about how precise a step match has to be, and closes the parked YAML-parser question in writing.

All three tasks landed; every checkbox on the issue's plan is ticked.

### Plan
- [x] Task 1: Anchor `StepWithId` on an exact `id:` match, and make ambiguity loud
- [x] Task 2: Start the slice at the step's own list-item header
- [x] Task 3: Add `Try*` variants and report a missing step uniformly

### What changed

Test infrastructure only — `FormCraft.UnitTests/Ci/`. No library code, no workflow, no `build/Build.cs`
behaviour change, and no new package (the parser stays rejected; see the issue's *Alternatives*).

- **`StepWithId` can no longer bind to the wrong step.** `id: <x>` was matched as an unanchored
  substring and `Array.FindIndex` took the first hit, so a search for `login` resolved to
  `id: login-legacy` — silently, on the primitive #226 depends on. It is now an exact trimmed-key
  match (trailing whitespace and a trailing `# comment` tolerated), and **two steps claiming one id
  fails loudly** instead of resolving to the first.
- **Both primitives return the same shape.** The slice starts at the step's own `- ` list item rather
  than at the `id:` line, so `StepWithId` now carries the step's `- name:` just as `StepNamed` does.
  An `id:` belonging to no list item reads as *absent* rather than slicing from line 0.
- **A missing step is reported once, by each test that cares.** New `TryStepNamed`/`TryStepWithId`
  return `string?`; the asserting forms are reimplemented over them with byte-identical messages, and
  `TestReportingTests`' five upload assertions route through the `Try*` form.

### Code review

A review over `origin/dev...HEAD` raised 14 findings; 9 were implemented in `fix: address
code-review findings`, and each fix was confirmed to correct a genuine wrong answer by running the old
and new algorithms side by side against the same fixtures:

| Finding | Old answer | Now |
|---|---|---|
| Walk-back had no containment check — a nested list item (`args:` entry, `path: \|` bullet) between the `id:` and its header stopped the search | returned a **fragment** of the step | walks past nested items to the step's own header |
| …and it could cross a **job boundary**, adopting the previous job's last step | returned `- name: real step` from another job | reads as absent (a shallower plain key ends the search) |
| `- ` matched literally — a bare `-` header, or `-  id:` | silently returned the **previous** step / reported a present step absent | `-` alone, `-<tab>` and `-  ` all recognised |
| `\s*(#.*)?$` let a longer id back in through the comment clause | `id: login#1` matched a search for `login` | comment requires preceding whitespace (`\s+#`) |
| `StepFrom`'s **end** boundary was not depth-aware either | truncated a step at its own nested sequence — and a truncated slice makes `ShouldNotContain` pass for the wrong reason | ends at the next *sibling* item |
| `StepWithId_Should_Treat_An_Id_Outside_Any_Step_As_Absent` passed **vacuously** — its fixture had no list item at all | false confidence on the one guard for the `-1` path | fixture now carries a real step in an earlier job |
| The `UploadStep` remark claimed "the count of red tests matches the count of causes" | **false** — 5 red before, 5 red after | states what actually changed: the messages, not the count |
| The absence rule was restated inline at 4 assertions | a 5th written `is { } step &&` would read a missing step as satisfied | centralised in one `UploadStepFails` helper |
| Two needless `!` null-forgiving operators | — | `return step.ShouldNotBeNull(…)` |

Also corrected two comments of mine that asserted things which are not true: step ids are unique per
**job**, not per workflow (so the new ambiguity failure is *not* evidence of an invalid workflow — the
real fix is `JobsOf` scoping, tracked on #198), and the unstripped-slice remark in
`TrustedPublishingWorkflowTests` claimed a harmlessness it does not have (see Follow-ups).

Five findings were **declined**, with reasons — three as explicit plan non-goals (job-scoping this
suite is #198; adding ambiguity detection to `StepNamed` and restructuring the guard set both change
*what* the suite asserts), one as plan-mandated surface (`TryStepWithId`'s throw-on-ambiguity is
deliberate and pinned by a test — returning `null` there would report a duplicated step as a missing
one), and one perf suggestion not worth trading precision for in a helper that runs seven times a run.

### Verification

- `./build.sh Test` (the exact CI target, invoked exactly as `ci.yml` does) — Restore/Compile/Test all
  succeeded, before and after the review fixes.
- `dotnet test -c Release` — **1351 passed, 1 skipped, 0 failed** (`FormCraft.UnitTests` 819, up from
  807 on `dev`; `ForMudBlazor` 464; `ForFluentUI` 68).
- **The reporting improvement was measured, not assumed.** With the `Publish: test-results` step
  temporarily deleted from `ci.yml`, on the pre-change code four of the five tests failed as
  `Shouldly.ShouldAssertException : Array.FindIndex(…) should be greater than or equal to 0 but was -1`
  — an apparent fault in the shared reader — while one named the offender. After the change all five
  report `[ci.yml / build-and-test]` as an offender collection, and the failure count is still exactly
  five, so no coverage was traded away. `ci.yml` was reverted; the workflow is untouched in this PR.

## Follow-ups

- **`ReleaseWorkflow_Should_Gate_The_Login_On_A_Created_Release` is inert, and pre-dates this PR.**
  It asserts `ShouldContain("release_created")` over the login step's *unstripped* slice, and that
  slice carries a comment block mentioning `release_created` in prose (`release-please.yml:153`). So
  deleting the step's `if:` outright — the exact #226 regression it exists to catch, whose cost is a
  green run on a version that never reached nuget.org — leaves it green. The prose sat inside the slice
  before this PR moved the slice's start, so this is not a regression here; it is left standing because
  narrowing it changes *what* the suite asserts, which #267 rules out. Fix is small: read
  `WorkflowSource.Stripped(...)` in that helper (the `# v1.2.0` digest annotation survives whole-line
  stripping, so the `uses: NuGet/login@<sha>` match is unaffected), or assert via `StepCondition` as
  the sibling #226 test already does. **Worth filing on its own.**
- **`StepNamed` still takes the first match, with no ambiguity check.** Step *names* need not be unique
  in GitHub Actions — unlike ids — so if a job ever splits its upload in two (a natural consequence of
  #256's per-project results directories), the four content assertions would read only the first and
  stay green while the second misbehaves. Declined here as a change to what the suite asserts.
- **The five absence branches have no automated coverage.** They were verified by hand (deleting the
  step from `ci.yml`, per the plan's own Step 6) but cannot execute in a normal run, because the guard
  set is discovered from the real workflows, which always contain the step. An inverted condition would
  ship green. Making the job set injectable — or a fixture overload of `UploadStep` — would close it.
- **`dotnet format` reports 57 pre-existing violations in `FormCraft.UnitTests`** — missing final
  newlines (~14 files), `IMPORTS` ordering in `GlobalUsings.cs`, and `IDE0011` missing braces in the
  validator/builder suites. **None are in the files this PR touches**, and CI does not gate on
  `dotnet format` (the enforcing gate is the Release build under `TreatWarningsAsErrors`, which is
  clean). Fixing them here would have meant a 50+ file diff unrelated to #267, so they are left for a
  deliberate formatting pass.

### DIFF
--- a/FormCraft.UnitTests/Ci/TestReportingTests.cs
+++ b/FormCraft.UnitTests/Ci/TestReportingTests.cs
@@ -190,13 +190,33 @@ public class TestReportingTests
     /// <remarks>
     /// Non-asserting since #267. The asserting form threw from in here, so deleting one upload step
     /// reddened five tests: the one whose actual subject is "every such job uploads" named the
-    /// offending job cleanly, and the other four died inside this helper — four extra failures
-    /// describing one root cause, presented as a fault in the shared reader rather than in the
-    /// workflow. Each caller below now treats absence as an offender of its own, so all five report
-    /// the same `<c>workflow / job</c>` pair and the count of red tests matches the count of causes.
+    /// offending job cleanly, and the other four died inside this helper, presented as a fault in the
+    /// shared reader rather than in the workflow.
+    /// <para>
+    /// What changed is the <em>message</em>, not the count. Measured on <c>ci.yml</c>: five tests were
+    /// red before and five are red after — but where four used to read
+    /// <c>Array.FindIndex(…) should be greater than or equal to 0 but was -1</c>, all five now name
+    /// <c>ci.yml / build-and-test</c> as an offender. The count cannot fall; every one of these claims
+    /// is genuinely unsatisfied by a job with no upload step.
+    /// </para>
     /// </remarks>
     private static string? UploadStep(TestRunningJob job) =>
         WorkflowSource.TryStepNamed(job.Text, UploadStepName);
+
+    /// <summary>
+    /// Whether <paramref name="job" />'s upload step fails <paramref name="claim" /> — including by
+    /// not existing at all.
+    /// </summary>
+    /// <remarks>
+    /// The absence rule lives here, once, rather than inline at each assertion below. Restating
+    /// <c>is not { } step ||</c> per call site puts a vacuity hole one forgotten clause away: written
+    /// as <c>is { } step &amp;&amp;</c>, a missing step reads as *satisfying* the claim, which is the
+    /// silence this file guards against everywhere else (see <see cref="JobsThatRunTests" />). Absence
+    /// offends every claim here — a step that is not there has no <c>if: always()</c>, and equally
+    /// does not "avoid the stale glob" in any sense worth being green about.
+    /// </remarks>
+    private static bool UploadStepFails(TestRunningJob job, Func<string, bool> claim) =>
+        UploadStep(job) is not { } step || !claim(step);
 
     [Fact]
     public void BuildScript_Should_Not_Set_The_VSTest_Properties_That_Mtp_Ignores()
@@ -347,10 +367,8 @@ public class TestReportingTests
         // one worth preserving, and it is exactly the path a bare step skips. Microsoft.Testing.
         // Platform prints only a summary line to stdout, so without this a red CI run leaves no
         // record of *which* assertion failed — a cost that was paid for real during #200.
-        // A job with no upload step at all offends this too — there is no `if: always()` on a step
-        // that is not there. Reported rather than thrown, so this test names the job like the others.
         var offenders = JobsThatRunTests()
-            .Where(j => UploadStep(j) is not { } step || !step.Contains("if: always()", StringComparison.Ordinal))
+            .Where(j => UploadStepFails(j, s => s.Contains("if: always()", StringComparison.Ordinal)))
             .ToList();
 
         offenders.ShouldBeEmpty();
@@ -365,8 +383,7 @@ public class TestReportingTests
         // re-introducing the very globs the next test rejects. All three steps upload exactly one
         // directory, so that is what is pinned.
         var offenders = JobsThatRunTests()
-            .Where(j => UploadStep(j) is not { } step
-                     || !step.Split('\n').Any(line => line.Trim() == "path: test-results"))
+            .Where(j => UploadStepFails(j, s => s.Split('\n').Any(line => line.Trim() == "path: test-results")))
             .ToList();
 
         offenders.ShouldBeEmpty();
@@ -380,12 +397,12 @@ public class TestReportingTests
         // created. --results-directory moved them, so that glob now matches nothing on every run —
         // the same declared-and-inert shape this issue exists to remove, just one file over. Every
         // target that tests routes through Test, so there is no run in which it could match again.
-        // Absence offends here too, even though a step that is not there points at nothing. Today a
-        // missing step makes this test red by throwing, so letting it read as vacuously satisfied
-        // would trade a loud failure for a quieter suite — the opposite of the point. The claim being
-        // kept is "every such job has an upload step, and it does not glob the stale path".
+        // The claim is inverted, so it reads "the step exists AND avoids the stale glob". Absence
+        // offends it, even though a step that is not there points at nothing: a missing step makes
+        // this test red today by throwing, so letting it read as vacuously satisfied would trade a
+        // loud failure for a quieter suite — the opposite of the point.
         var offenders = JobsThatRunTests()
-            .Where(j => UploadStep(j) is not { } step || step.Contains("**/TestResults/", StringComparison.Ordinal))
+            .Where(j => UploadStepFails(j, s => !s.Contains("**/TestResults/", StringComparison.Ordinal)))
             .ToList();
 
         offenders.ShouldBeEmpty();
@@ -400,8 +417,7 @@ public class TestReportingTests
         // real one gets missed. The opposite risk — an *empty* directory passing unremarked — is
         // covered in the build rather than here, by the Assert.NotEmpty on the trx.
         var offenders = JobsThatRunTests()
-            .Where(j => UploadStep(j) is not { } step
-                     || !step.Contains("if-no-files-found: ignore", StringComparison.Ordinal))
+            .Where(j => UploadStepFails(j, s => s.Contains("if-no-files-found: ignore", StringComparison.Ordinal)))
             .ToList();
 
         offenders.ShouldBeEmpty();
--- a/FormCraft.UnitTests/Ci/TestReportingTests.cs
+++ b/FormCraft.UnitTests/Ci/TestReportingTests.cs
@@ -184,10 +184,19 @@ public class TestReportingTests
 
     /// <summary>
     /// The upload step as it appears <em>within that job</em> — so an assertion about it cannot be
-    /// answered by an identically-named step sitting in a sibling job.
+    /// answered by an identically-named step sitting in a sibling job — or <c>null</c> when that job
+    /// has no such step.
     /// </summary>
-    private static string UploadStep(TestRunningJob job) =>
-        WorkflowSource.StepNamed(job.Text, UploadStepName, job.ToString());
+    /// <remarks>
+    /// Non-asserting since #267. The asserting form threw from in here, so deleting one upload step
+    /// reddened five tests: the one whose actual subject is "every such job uploads" named the
+    /// offending job cleanly, and the other four died inside this helper — four extra failures
+    /// describing one root cause, presented as a fault in the shared reader rather than in the
+    /// workflow. Each caller below now treats absence as an offender of its own, so all five report
+    /// the same `<c>workflow / job</c>` pair and the count of red tests matches the count of causes.
+    /// </remarks>
+    private static string? UploadStep(TestRunningJob job) =>
+        WorkflowSource.TryStepNamed(job.Text, UploadStepName);
 
     [Fact]
     public void BuildScript_Should_Not_Set_The_VSTest_Properties_That_Mtp_Ignores()
@@ -320,8 +329,12 @@ public class TestReportingTests
     {
         // In that same job, not merely somewhere in the file: the artifact is uploaded from the
         // workspace the build filled, and a sibling job's workspace is a different one.
+        //
+        // Asked through the same primitive as the four tests below rather than by a raw substring
+        // search, so "present" means the one thing here and there — a step the scan can actually
+        // isolate, not merely the text of a `- name:` line appearing somewhere in the job.
         var missing = JobsThatRunTests()
-            .Where(j => !j.Text.Contains($"- name: '{UploadStepName}'", StringComparison.Ordinal))
+            .Where(j => UploadStep(j) is null)
             .ToList();
 
         missing.ShouldBeEmpty();
@@ -334,8 +347,10 @@ public class TestReportingTests
         // one worth preserving, and it is exactly the path a bare step skips. Microsoft.Testing.
         // Platform prints only a summary line to stdout, so without this a red CI run leaves no
         // record of *which* assertion failed — a cost that was paid for real during #200.
+        // A job with no upload step at all offends this too — there is no `if: always()` on a step
+        // that is not there. Reported rather than thrown, so this test names the job like the others.
         var offenders = JobsThatRunTests()
-            .Where(j => !UploadStep(j).Contains("if: always()", StringComparison.Ordinal))
+            .Where(j => UploadStep(j) is not { } step || !step.Contains("if: always()", StringComparison.Ordinal))
             .ToList();
 
         offenders.ShouldBeEmpty();
@@ -350,9 +365,8 @@ public class TestReportingTests
         // re-introducing the very globs the next test rejects. All three steps upload exactly one
         // directory, so that is what is pinned.
         var offenders = JobsThatRunTests()
-            .Where(j => !UploadStep(j)
-                .Split('\n')
-                .Any(line => line.Trim() == "path: test-results"))
+            .Where(j => UploadStep(j) is not { } step
+                     || !step.Split('\n').Any(line => line.Trim() == "path: test-results"))
             .ToList();
 
         offenders.ShouldBeEmpty();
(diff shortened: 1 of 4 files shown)


phmatray:
Noted while merging — **recorded, not filed** (no action planned):

`dotnet format FormCraft.UnitTests/FormCraft.UnitTests.csproj --verify-no-changes` reports **57 pre-existing violations**: missing final newlines (~14 files), `IMPORTS` ordering in `GlobalUsings.cs`, and `IDE0011` missing braces across the validator/builder suites. **None are in the files this PR touched**, and CI does not gate on `dotnet format` — the enforcing gate is the Release build under `TreatWarningsAsErrors`, which is clean.

Deliberately **not** filed as an issue: the fix is a single `dotnet format FormCraft.sln` run touching 50+ files, and with several PRs in flight against `dev` right now that diff would conflict with all of them. Better done as a deliberate, quiet pass when the queue is empty — this comment is the record so it can be found then.

PR #290, opened 2026-08-13
docs(profile): document the working per-class test filter (#277)

### LINKED ISSUE #277: The repo profile says per-suite test filtering is impossible; the MTP runner supports it
## Problem / motivation

`.claude/skills/repo-profile.md` — the file the issue/PR lifecycle skills read for repo facts — records
per-suite test filtering as impossible:

> **Single-suite filter (per-task, fast):** **unavailable.** `dotnet test --filter …` is *silently
> ignored* here — the projects use Microsoft.Testing.Platform, which emits `MTP0001` and runs the whole
> suite anyway. Always run the full suite; never trust a filtered run's scope.

and again under *Environment gotchas*:

> **`dotnet test --filter` does nothing** (see *Build & test*) — plans that prescribe a per-task filter
> must run the full suite instead.

**The first sentence is true; the conclusion drawn from it is not.** `--filter` is a VSTest option, and
`dotnet test` forwards it as an MSBuild property that Microsoft.Testing.Platform ignores — that part is
correct, and is the same class of inert wiring #231 removed. But the *runner* is not the shim. Both test
projects set `<OutputType>Exe</OutputType>` and `<UseMicrosoftTestingPlatformRunner>true</…>`, so the
built assembly is an MTP host with its own command line, and xunit.v3's runner exposes real filters on
it:

```console
$ FormCraft.UnitTests/bin/Release/net10.0/FormCraft.UnitTests --filter-class FormCraft.UnitTests.Ci.GitignoreTests
Test run summary: Passed! - …/FormCraft.UnitTests.dll (net10.0|arm64)
```

`--filter-class`, `--filter-method`, `--filter-namespace`, their `--filter-not-*` counterparts and
`--filter-uid` are all present (`--help` lists them; `*` wildcards are supported at either end).

**Why it matters.** The profile's advice is load-bearing for every agent-driven task loop — plans are
explicitly told to run the whole suite because filtering "is unavailable". That is ~1200 tests and a
full restore/build per check. Measured while building #271: a full `dotnet test -c Release` cycle is
roughly a minute, while the same class run through the built binary returns in well under a second.
That PR needed a **10-case verification matrix** (a baseline plus nine deliberately broken
`.gitignore` variants); at full-suite cost that is ~10 minutes of rebuild-and-rerun, which is enough
friction that the honest version of the check does not get written.

### The trap that makes this worth writing down

Passing the *wrong* flag to the MTP host does not fail:

```console
$ FormCraft.UnitTests/bin/Release/net10.0/FormCraft.UnitTests --filter "*GitignoreTests*"
… prints the full --help text …
$ echo $?
0
```

No test runs, nothing is reported, and the exit code is **0**. A verification loop that greps for
`Failed!` sees none and reads it as a pass — which is exactly what happened on the first attempt at
#271's matrix: nine traps all reported "no result", and only an explicit check of the raw output
revealed that nothing had executed. Any correction to the profile must name this, or it trades one
silent-inertness trap for another.

## Proposed solution

Correct both places in `.claude/skills/repo-profile.md`:

1. Replace *"Single-suite filter: **unavailable**"* with the working invocation — the built binary plus
   `--filter-class` / `--filter-method` / `--filter-namespace` — while **keeping** the existing warning
   that `dotnet test --filter` is inert, since that remains true and is the trap people arrive with.
2. Add the exit-0-on-bad-flag warning, with the instruction to assert on a real `Passed!` / `Failed!`
   summary line rather than on the exit code or on the absence of a failure string.
3. Note the prerequisite: the binary must be built first (`dotnet build -c Release`), so this is a
   *within-task iteration* tool, not a replacement for the profile's *CI gates* — `./build.cmd Test`
   stays the pre-merge gate and still runs everything.

Also verify, and record, whether `dotnet test -- --filter-class …` (SDK argument passthrough to the MTP
host) works on the pinned SDK. If it does it is the more ergonomic form and should be the one the
profile leads with; if it does not, say so explicitly so nobody retries it.

## Alternatives considered

- **Leave the profile as-is.** The advice is merely conservative, not dangerous — running everything is
  always correct. Rejected: it is wrong in a way that costs real time on every task, and being wrong in
  the file agents treat as ground truth is worse than being wrong in a comment.
- **Only add the flags, drop the `dotnet test --filter` warning.** Shorter. Rejected: the warning is why
  the misconception exists, and `MTP0001` is genuinely silent — deleting the explanation invites the
  next person to "fix" the profile back.
- **Add a guard test asserting the flags still work.** Tempting given `FormCraft.UnitTests/Ci/` exists for
  repo invariants. Rejected as out of scope: a test that shells out to the test binary to test the test
  binary is circular, slow, and would break on any xunit.v3 runner rename — the failure mode here is a
  stale doc, not a silent regression, and a stale doc is caught by the next person reading it.

## Area

Repository tooling — `.claude/skills/repo-profile.md`, test invocation

---

**Follow-up from #271** (which measured this while guarding `.gitignore`). **Related:** #264, #231, #256

<details>

### Problem / context

The repo profile is consumed by `create-issue`, `implement-issue` and `merge-pr` as the authority on how
to build and test this repository. One of its claims — that no per-suite filter exists — is a wrong
inference from a correct observation, and it is the claim with the highest per-run cost, because it
appears in every generated plan's Global Constraints ("`dotnet test --filter` is inert (`MTP0001`) — run
the whole suite").

The correction is small but has to be *precise*, because the neighbourhood is full of things that look
like they work and do not: `--filter` (inert via the shim), `--filter` on the host (prints help, exits 0),
and `VSTestLogger`/`VSTestResultsDirectory` (the inert wiring #231 removed). A vague correction here
produces another one of those.

### Approaches

**A. Correct both profile sections, name the exit-0 trap, and check the `dotnet test --` passthrough.**
*Pros:* fixes the wrong claim, keeps the true half, and inoculates against the failure mode that the
correction itself introduces. *Cons:* requires one measurement (the passthrough form) before writing.

**B. Correct the flags only.** *Pros:* a two-line edit. *Cons:* leaves the exit-0 trap undocumented —
and that trap is more dangerous than the original error, because it produces a *false green* rather
than merely slow-but-correct behaviour.

**C. Correct the profile and add a `Ci/` guard test.** *Pros:* the invariant stops depending on a human
reading it. *Cons:* circular (a test asserting the test runner's CLI), slow, and brittle against runner
renames — for a doc whose readers are agents that re-read it every run anyway.

### Recommendation

**A.** B's omission converts a performance problem into a correctness one, which is the wrong trade for
two saved lines. C is the kind of guard that is right for build wiring (#231's `Assert.NotEmpty`, where
nothing human ever looks) and wrong for a doc that is read aloud at the start of every task.

</details>

<details>

### Goal

`.claude/skills/repo-profile.md` describes test invocation accurately: the fast path exists and is
spelled correctly, the inert path is still flagged, and the false-green trap is called out.

### Scope

- Edit the *Build & test* → "Single-suite filter" bullet.
- Edit the *Environment gotchas* → "`dotnet test --filter` does nothing" bullet.
- Measure and record whether `dotnet test -- --filter-class …` works on the pinned SDK (`global.json`
  → 10.0.301).

### Non-goals

- Changing any build script, CI workflow, or `.csproj`. `./build.cmd Test` remains the CI gate and
  keeps running the full suite — this is about local/agent iteration only.
- Adding a test that asserts the runner's CLI surface (rejected as approach C).
- Auditing the rest of the profile. (Its own header already warns that fields measured from a stale
  branch can be wrong; a full re-verify is separate work.)

### Behaviour

```mermaid
flowchart TD
    A["dotnet test --filter X"] --> B["shim forwards as MSBuild property"]
    B --> C["MTP ignores it — MTP0001<br/>WHOLE SUITE RUNS"]
    D["&lt;binary&gt; --filter X"] --> E["unknown option<br/>prints --help, EXIT 0<br/>NOTHING RUNS — reads as a pass"]
    F["&lt;binary&gt; --filter-class X"] --> G["runs just that class<br/>real Passed!/Failed! summary"]
```

### Key files

- `.claude/skills/repo-profile.md` — the only file changed.
- `FormCraft.UnitTests/FormCraft.UnitTests.csproj` / `FormCraft.ForMudBlazor.UnitTests.csproj` — evidence
  for the claim (`OutputType=Exe`, `UseMicrosoftTestingPlatformRunner=true`); not modified.

### Validation rules

- The documented `--filter-class` invocation runs a strict subset of the suite and prints a real
  `Passed!` / `Failed!` summary line.
- The documented warning about bare `--filter` matches observed behaviour (help text, exit 0).
- Both edited bullets still state that `dotnet test --filter` is inert.

### Edge cases

- **Exit code is not a verdict here.** Assert on the summary line; a bad flag exits 0 having run nothing.
- **The binary must exist.** Path is configuration- and TFM-dependent
  (`bin/<cfg>/net10.0/FormCraft.UnitTests`); a stale binary silently tests old code, so a build must
  precede it.
- Test projects are single-target `net10.0`, so there is one binary per project — no per-TFM ambiguity.

### Assumptions

- Both test projects keep the MTP runner; if either reverts to VSTest, this guidance changes with it.
- Recorded from a `dev`-based checkout, per the profile header's own warning about branch-dependent
  measurements.

</details>

## 🛠️ Implementation plan

**Goal:** make the profile's test-invocation guidance true, fast, and trap-aware.

**Architecture:** documentation only — one Markdown file. No library, build-script or workflow change.

**Tech stack:** .NET 10 SDK pinned by `global.json` (10.0.301), Microsoft.Testing.Platform + xunit.v3.

**Global constraints:**
- ⛔ Do **not** change `build/Build.cs`, any `.github/workflows/*`, or any `.csproj` — `./build.cmd Test`
  stays the CI gate and stays full-suite.
- Every documented command must be **run** before it is written down; this issue exists because a
  plausible-sounding claim about this exact tooling went unmeasured.
- Measure from a `dev`-based checkout (the profile header's own warning).

### Task 1: Measure the real invocation surface

**Files:** none modified — this task produces the evidence Task 2 writes down.

**Interfaces:** none.

- [x] **Step 1:** Build once: `dotnet build -c Release`.
- [x] **Step 2:** Run `FormCraft.UnitTests/bin/Release/net10.0/FormCraft.UnitTests --filter-class FormCraft.UnitTests.Ci.GitignoreTests` and record the summary line and the test count — confirm it is a strict subset of the full suite, not all ~805.
- [x] **Step 3:** Run the same binary with bare `--filter "*Gitignore*"`; record that it prints help and record `echo $?` (expected `0`, running nothing). This is the trap the docs must name.
- [x] **Step 4:** Run `dotnet test FormCraft.UnitTests/FormCraft.UnitTests.csproj -c Release -- --filter-class FormCraft.UnitTests.Ci.GitignoreTests` and record whether the passthrough reaches the host (subset) or is ignored/rejected — this decides which form Task 2 leads with.
- [x] **Step 5:** Repeat Step 2 against `FormCraft.ForMudBlazor.UnitTests` to confirm the guidance is not project-specific.
- [x] **Step 6:** Commit: `docs(profile): record the measured MTP filter invocations`.

### Task 2: Correct the profile

**Files:** modify `.claude/skills/repo-profile.md` (the *Build & test* and *Environment gotchas* bullets).

**Interfaces:** none.

- [x] **Step 1:** Rewrite the *Build & test* → "Single-suite filter" bullet: lead with the form Task 1 Step 4 proved, give a copy-pasteable example, keep the `MTP0001` explanation for `dotnet test --filter`, and note that a build must precede the binary run.
- [x] **Step 2:** Rewrite the *Environment gotchas* → "`dotnet test --filter` does nothing" bullet to point at the working form instead of at "run the full suite".
- [x] **Step 3:** Add the exit-0 warning explicitly: a bad flag prints `--help` and exits `0`, so assert on the `Passed!`/`Failed!` summary line, never on the exit code or on the absence of a failure string.
- [x] **Step 4:** State the scope boundary — this is for local/agent iteration; `./build.cmd Test` remains the CI gate and still runs the whole suite.
- [x] **Step 5:** Re-run every command now written in the file, verbatim as written, and confirm each behaves as documented (copy-paste errors in a doc of exact commands are the whole failure mode here).
- [x] **Step 6:** Run `dotnet build -c Release` to confirm the repo still builds clean under `TreatWarningsAsErrors` (no code changed, but the gate is cheap and the profile claims it).
- [x] **Step 7:** Commit: `docs(profile): document the working per-class test filter`.

phmatray:
Implements #277.

Closes #277.

The repo profile records per-suite test filtering as impossible. The premise is right — `dotnet test --filter` really is inert under Microsoft.Testing.Platform (`MTP0001`) — but the conclusion isn't: the built test assemblies are MTP hosts with their own command line, and xunit.v3's runner exposes real `--filter-class` / `--filter-method` / `--filter-namespace` options on them.

Correcting both bullets, and naming the trap the correction itself would otherwise introduce: passing the *wrong* flag to the host prints `--help`, runs nothing, and exits **0** — a false green for any loop that greps for `Failed!`.

code-review pass.

### Plan
- [x] Task 1: Measure the real invocation surface
- [x] Task 2: Correct the profile

### What was measured

Every command in the new text was run before it was written down, and re-run verbatim afterwards
(SDK `10.0.302`, xunit.v3 runner `3.2.2`, Release):

| Invocation | Result |
|---|---|
| `dotnet test <csproj> -c Release -- --filter-class <FQN>` | **works** — `Total: 6`, exit 0, ~10s incl. build → the form the profile leads with |
| `<binary> --filter-class <FQN>` | **works** — `total: 6` of 808, 191 ms |
| `<binary> --filter-namespace FormCraft.UnitTests.Ci` | `total: 59` of 808 |
| `<binary> --filter-class '*GitignoreTests'` | `total: 6` — wildcards work |
| MudBlazor host, `--filter-class …AriaRequiredTests` | `total: 34` of 464 — not project-specific |
| `dotnet test --filter FullyQualifiedName~…` | **still inert** — `MTP0001`, `Total: 808`, whole suite ran |
| `<binary> --filter "*Gitignore*"` | `Unknown option`, full `--help`, **no summary line**, direct exit **5** |
| `<binary> --filter-method '<bare name>'` | **`Zero tests ran`, exit 8** — needs the fully-qualified name |
| `./build.sh Test` | 1340 tests (808 + 464 + 68), green, ~30s |

Two corrections to the issue's own premises, both measured rather than assumed:

1. **The bad flag exits `5`, not `0`.** The false green is real but the mechanism is different — piping
   (`| tail`, `| grep`, as agents habitually do) replaces `$?` with the pipe's `0`, and the run emits
   *no summary line at all*, so `grep 'Failed!'` matches nothing. The profile now says to assert on the
   summary line and warns that the pipe, not the binary, manufactures the pass.
2. **A second trap, not in the issue:** `--filter-method` silently matches nothing when given a bare
   method name — it wants `<namespace>.<class>.<method>`, or a wildcard. That prints `Zero tests ran`
   and exits `8`.

### Code review

A review over `dev...HEAD` returned 12 findings, all independently re-measured before acting. Nine
were real and are fixed in `fix: address code-review findings`; the substantive ones:

- **The two paths report failure differently** — the biggest gap. Via the *Preferred* passthrough a
  zero-match prints `Failed! … Total: 0` + exit 1 and **never says `Zero tests ran`**, and a wrong
  flag's `Unknown option` diagnostic never reaches stdout at all (it goes to a `TestResults` log,
  while stdout says `error run failed: Tests failed`). The original text described only the
  direct-binary behaviour, so its verification rule did not apply to the form it recommended.
- **Solution-level passthrough is a live trap.** `dotnet test -c Release -- --filter-class X`
  applies the filter to all three assemblies → one `Passed! … Total: 6` plus two
  `Failed! … Total: 0`, exit 1. Grepping `Passed!` reports green on a failed command. The example
  now names its `.csproj` and the trap is documented.
- **Self-contradiction removed:** the text claimed a bare `--filter-method` name "matches nothing
  and does not say so" while elsewhere documenting `Zero tests ran` — it does say so.
- **Fabricated output line** `# total: 6 (of 808), 191ms` replaced with the real two-line shape, so
  the "grep the summary line" rule is actually executable.
- **No `1340` aggregate is ever printed** — three per-assembly summaries in nondeterministic order;
  a check reading the first accepts `Total: 68` as the whole suite.
- **Duplicate guidance collapsed:** the *Environment gotchas* entry is a pointer again, not a second
  copy that would drift.
- Also added: simple filters cannot be combined with `--filter-query`; the SDK stamp now comes from
  `dotnet --version` rather than the `global.json` pin (`rollForward: latestFeature` means the pin
  cannot signal drift); Windows `.exe` caveat, marked as not measured here.

**Scope call:** the review also flagged the **Worktree home** section as flatly false — it claimed
`.claude/worktrees/` was unignored due to a collapsed `.gitignore:13`. Verified: line 13 is
`/test-results/`, line 16 is `.claude/worktrees/`, split apart by #223 (`6e8a10c`) and pinned by the
very `GitignoreTests` class this PR uses as its example. Although "auditing the rest of the profile"
is a stated non-goal, leaving a **proven-false** claim in the one file this PR exists to make
truthful would be self-defeating, so it is corrected here. Flagged rather than smuggled.

## Follow-ups

Deliberately **not** fixed here — genuinely out of scope (different file / different audience):

- **`CLAUDE.md` still teaches five inert `dotnet test --filter …` invocations** (lines 37-43),
  including `--filter "Category=Builder|Renderer|Security"`. Measured: `--filter` is ignored
  (`MTP0001`, `Total: 808` — the whole suite), *and* the repo contains zero `[Trait]` attributes, so
  the `Category` examples never matched anything. This matters more than the profile: `CLAUDE.md` is
  auto-loaded every session, while `repo-profile.md` is read only by the lifecycle skills.
- `CLAUDE.md` also still says "600+ unit tests across 2 test projects" — it is now **1340** across
  **three** (`FormCraft.ForFluentUI.UnitTests`, added in #261).
- The alternate worktree home `.worktrees/` is not ignored (`worktrees-ignored.sh` exits `1`).
  Now noted in the profile; the one-line `.gitignore` addition was not taken unasked.
- Minor: the issue's own spec cites `global.json` as SDK `10.0.301`; it pins `10.0.302`.

### DIFF
--- a/.claude/skills/repo-profile.md
+++ b/.claude/skills/repo-profile.md
@@ -18,33 +18,45 @@
 - **Single-suite filter (per-task, fast):** **available — but never via `dotnet test --filter`.**
   The test projects are Microsoft.Testing.Platform hosts (`OutputType=Exe` +
   `UseMicrosoftTestingPlatformRunner=true`), so xunit.v3's runner exposes real filter options on
-  them. Both forms below run a strict subset and print a genuine summary line. Measured on SDK
-  `10.0.302` / xunit.v3 runner `3.2.2` (#277) — re-measure if either moves.
-  - **Preferred** (builds first; the SDK forwards everything after `--` to the host):
+  them. Both forms below run a strict subset. ⚠️ **They report *failure* differently — read the
+  path-specific rules below before writing a verification loop.** Recorded on macOS arm64,
+  `dotnet --version` = `10.0.302`, xunit.v3 runner `3.2.2` (#277). Re-stamp from `dotnet --version`,
+  **not** from `global.json`: `rollForward: latestFeature` lets the real SDK drift above the pin, so
+  the pinned string can never tell you the measurement went stale.
+  - **Preferred** (builds first; the SDK forwards everything after `--` to the host).
+    ⚠️ **Always name the `.csproj`** — see the solution-level trap below:
     ```bash
     dotnet test FormCraft.UnitTests/FormCraft.UnitTests.csproj -c Release \
       -- --filter-class FormCraft.UnitTests.Ci.GitignoreTests
-    # Passed! - Failed: 0, Passed: 6, Skipped: 0, Total: 6 — ~10s including the build
     ```
+    → `Passed! - Failed: 0, Passed: 6, Skipped: 0, Total: 6 … - FormCraft.UnitTests.dll`, exit `0`,
+    ~10s including the build.
   - **Fastest** (runs the built host directly — **it does not build**, so `dotnet build -c Release`
-    must precede it or you are silently testing stale code):
+    must precede it or you are silently testing stale code). Path is relative to the repo root, and
+    is `…\FormCraft.UnitTests.exe` on Windows (not measured here — recorded on macOS):
     ```bash
     FormCraft.UnitTests/bin/Release/net10.0/FormCraft.UnitTests \
       --filter-class FormCraft.UnitTests.Ci.GitignoreTests
-    # total: 6 (of 808), 191ms
     ```
+    → `Test run summary: Passed! - …/FormCraft.UnitTests.dll (net10.0|arm64)` on one line, then
+    `  total: 6` / `  failed: 0` / `  succeeded: 6` / `  skipped: 0` / `  duration: 191ms` on the
+    next five. **The verdict and the count are on different lines** — one `grep` gets you one of
+    them, never both.
     Not project-specific — every test project has its own host:
     ```bash
     FormCraft.ForMudBlazor.UnitTests/bin/Release/net10.0/FormCraft.ForMudBlazor.UnitTests \
       --filter-class FormCraft.ForMudBlazor.UnitTests.Fields.AriaRequiredTests
-    # total: 34 (of 464)
     ```
+    → `Passed!` with `total: 34` (of 464).
   - **Flags** (`--help` on the host lists them all): `--filter-class`, `--filter-method`,
     `--filter-namespace`, `--filter-trait`, `--filter-uid`, `--filter-query`, plus a
     `--filter-not-*` counterpart for class/method/namespace/trait. `*` wildcards work at either end
     (`--filter-class '*GitignoreTests'` → 6).
-    - ⚠️ **`--filter-method` wants the fully-qualified name** (`<namespace>.<class>.<method>`) —
-      a bare method name matches **nothing** and does not say so; a wildcard
+    Simple filters (`--filter-class/-method/-namespace/-trait` and their `-not-` forms) **cannot be
+    combined with `--filter-query`** — the runner's own help says so.
+    - ⚠️ **`--filter-method` wants the fully-qualified name** (`<namespace>.<class>.<method>`, as
+      its `--help` states) — a bare method name matches nothing and reports that as
+      `Zero tests ran` (exit `8`), which is easy to skim past as success. A wildcard
       (`--filter-method '*Clearing_A_Standalone*'`) is the ergonomic escape.
     - ⚠️ **`--filter-trait` is dead weight here:** no test in this repo carries a `[Trait]`, so
       `--filter-trait 'Category=Builder'` returns `Zero tests ran`. Note `CLAUDE.md` still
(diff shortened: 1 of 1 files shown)


phmatray:
Follow-up triage for this merge (per `/merge-pr` Step 6):

**Filed as an issue** — #299, *"CLAUDE.md still teaches VSTest test commands that Microsoft.Testing.Platform silently ignores"*. The two `CLAUDE.md` findings from this PR's Follow-ups section clustered into one root cause, plus a third instance found while triaging: `dotnet test --collect:"XPlat Code Coverage"` (CLAUDE.md:34) also emits `MTP0001 … VSTestCollect` and produces **no coverage file at all**, while exiting `0`.

**Recorded here rather than filed** — both are observations, not committed work:

- **`.worktrees/` is not git-ignored** in this repo (the kit's *other* default worktree home; `worktrees-ignored.sh` exits `1` for it). Nothing currently plants worktrees there — `.claude/worktrees/` is the home in use and *is* ignored — so this is latent, not active. It is now noted in `.claude/skills/repo-profile.md`; a one-line `.gitignore` addition would settle it if it ever becomes real. Not taken unasked.
- **#277's spec cited `global.json` as SDK `10.0.301`**; it actually pins `10.0.302`, which the profile already had right. Noted for the record — the issue is closed and nothing depends on it.

No other deferred work was found in the PR body or review comments.

Issue #279, opened 2026-08-12
Move the UI-agnostic adapter machinery into FormCraft core


phmatray:
## Problem / motivation

`FormCraft` (core) advertises itself as UI-framework-agnostic, and until #260 that claim had never
been tested — there was exactly one adapter, so nothing could reveal where the boundary was actually
drawn. Building `FormCraft.ForFluentUI` tested it, and found the boundary in the wrong place in three
distinct ways.

**1. UI-agnostic code lives in the MudBlazor package, so the second adapter had to duplicate it.**
`FormCraft.ForMudBlazor.DynamicFormValidator<TModel>` is 242 lines that reference **nothing** from
MudBlazor — only `Microsoft.AspNetCore.Components`. It is the component that makes
`.Required(...)` and `.WithValidator(...)` produce actual validation messages. Because it is in that
package, `FormCraft.ForFluentUI` had to ship its own `FluentUIDynamicFormValidator<TModel>` covering
the non-collection half, with an XML comment admitting the duplication and asking for this issue. The
same applies to two smaller pieces:

| Member | Where it lives | Why the Fluent adapter could not use it |
|---|---|---|
| `DynamicFormValidator<TModel>` | `FormCraft.ForMudBlazor` | wrong package; copying it was the only option that did not break the MudBlazor package's public API |
| `NativeRequired.Resolve(...)` | `FormCraft.ForMudBlazor` (`internal`) | `internal`, so the rule was hand-reimplemented in `FluentUIFieldComponentBase.EffectiveNativeRequired` |
| `WithNativeRequired(...)` | `FormCraft.ForMudBlazor` | Fluent users must write the raw `.WithAttribute("Required", …)` form instead |

`NativeRequired`'s own XML doc calls itself *"the single implementation of the native-required rule
(#199)… so `RenderPipelineParityTests` is guarding against regressions rather than against a
copy-paste drifting."* It is now copy-pasted across two packages — precisely the drift it was written
to prevent.

**2. The seam the docs describe does not exist.** `CLAUDE.md` presents `IUIFrameworkAdapter` as the
adapter contract, with this shape:

```csharp
public interface IUIFrameworkAdapter
{
    RenderFragment RenderField<TModel>(IFieldRenderContext<TModel> context);
    RenderFragment RenderForm<TModel>(IFormConfiguration<TModel> config);
}
```

The real interface has neither method. More importantly, a solution-wide reference search returns
[...]

Issue #260, opened 2026-08-12
Add a FormCraft.ForFluentUI adapter for Fluent UI Blazor v5


phmatray:
## Problem / motivation

FormCraft ships one UI adapter: `FormCraft.ForMudBlazor`. Teams standardised on Microsoft's design
language — Fluent — have no way to use this library without adopting Material Design, which for a
line-of-business app inside the Microsoft ecosystem is usually a non-starter.

There is a second, structural motivation. `FormCraft` (core) advertises itself as UI-framework-agnostic,
and `CLAUDE.md` documents an `IUIFrameworkAdapter` seam for exactly this purpose. That claim has never
been exercised: with a single adapter in the tree, nothing proves the seam is real. Building a second
adapter is the only way to find out — and the investigation behind this issue already found that the
advertised seam is **not** the live one (see the Brainstorm).

Fluent UI Blazor is the natural second target: it is Microsoft-published, MIT-licensed, and v5 —
a full rewrite onto Fluent UI Web Components v3 — is at `5.0.0-rc.5-26219.1` with a stable API
surface. v4 is supported only until **November 2026**, so any adapter written against v4 would be
born deprecated. This issue targets v5 directly.

## Proposed solution

Add a new package, **`FormCraft.ForFluentUI`**, that plugs into the same live extension seam
`FormCraft.ForMudBlazor` uses — `FieldRendererBase` subclasses registered in a precedence-ordered DI
extension — and provides a Fluent form container plus the **five type-based field renderers**: text,
numeric (including nullable), boolean, date/time, and select.

Swapping adapters becomes a two-line change for a consumer, because the Fluent container keeps the
same type name in its own namespace:

```csharp
// Program.cs
builder.Services.AddFormCraft();
builder.Services.AddFormCraftFluentUI();   // was: AddFormCraftMudBlazor()

// _Imports.razor
@using FormCraft.ForFluentUI              // was: @using FormCraft.ForMudBlazor
```

The two adapters are **mutually exclusive by design**, and `AddFormCraftFluentUI()` throws if the
MudBlazor renderers are already registered — renderer selection is first-match-wins, so a container
holding both would silently render a half-Material, half-Fluent form.

Deliberately **out of scope**, each tracked as a follow-up rather than smuggled in:

- **Collection / item-form fields** (the imperative `RenderTreeBuilder` path) — blocked on #203.
- Lookup and LOV dialogs, autocomplete, multi-select, file upload.
[...]

PR #291, opened 2026-08-13
feat(fluentui): bring the Fluent UI adapter to parity with MudBlazor (#278)

### LINKED ISSUE #278: Bring FormCraft.ForFluentUI to parity with the MudBlazor adapter
## Problem / motivation

`FormCraft.ForFluentUI` shipped in #260 as a deliberate vertical slice: a form container plus the five
type-based field types (text, numeric, boolean, date/time, select). Everything else the MudBlazor
adapter does is absent, and the gap is wider than "some field types are missing".

Three of the gaps are not merely missing features — they change what a *working* configuration does
when you switch adapters:

- **`.WithSecurity(...)` is not enforced.** Rate limiting, CSRF protection, audit logging and field
  encryption are all implemented in `FormCraft.ForMudBlazor`'s container and in none of the Fluent
  one. #260 shipped a fail-closed guard — the Fluent container throws `NotSupportedException` at
  render time rather than rendering a form that *looks* protected — so nothing is silently unsafe
  today, but a Fluent user simply cannot build a secured form.
- **Collection / item-form fields do not render at all.** This was blocked on #203 when #260 was
  planned. #203 has since landed, so item fields now flow through `IFieldRendererService` like every
  other field — which means the Fluent adapter gets them by implementing renderers, not by
  reimplementing a second imperative `RenderTreeBuilder` path. **The blocker is gone.**
- **Field groups render, but ungrouped.** `.AddFieldGroup(...)` fields appear as a flat list; the
  card, column and collapsible layout the MudBlazor container gives them is missing, so a form
  designed around grouping loses its structure rather than reporting a problem.

The remaining gaps are ordinary missing field types: lookup and LOV dialogs, autocomplete,
multi-select, file upload, and the three custom renderers (colour picker, rating, slider).

## Proposed solution

Close the gap in **priority order, not alphabetically** — the ordering is the substance of this issue,
because the items are not equally valuable and the list is long enough that starting at the wrong end
wastes the effort:

1. **`.WithSecurity(...)` enforcement** — the only item where the current behaviour blocks a whole
   class of application outright. Replaces the `NotSupportedException` guard.
2. **Collection / item-form fields** — now cheap, since #203 made item fields go through the ordinary
   renderer registry.
3. **Field-group layout** — affects forms that already work, and is contained to the container.
4. **Remaining field types** — lookup, LOV, autocomplete, multi-select, file upload.
5. **Custom renderers and the demo page** — colour picker, rating, slider; a Fluent showcase page.
6. **A typed `.WithNativeRequired(...)` for Fluent** — depends on where the builder extension ends up
   living, which is the subject of the sibling issue below.

Each is independently shippable, so this can land over several PRs rather than one.

## Alternatives considered

- **One big "achieve parity" PR.** Rejected: #260 was already large as a five-field-type slice, and a
  PR spanning security enforcement, dialogs, file upload and the demo app would be unreviewable.
- **Declare the Fluent adapter feature-frozen at its current slice.** Defensible if the adapter were
  an experiment, but #260 published it to NuGet as a real package with a documented "not yet covered"
  list — which is a promise to close it, not a permanent boundary.
- **Do the field types first and security last** (the natural order, since field types are easier).
  Rejected: security is the only item that currently makes the adapter unusable rather than merely
  incomplete, and leaving it last means the throw stays in place longest.

## Area

Rendering / UI framework adapters — `FormCraft.ForFluentUI`.

**Related:** #260, #203, #250

---

<details>

### Problem & context

`FormCraft.ForFluentUI` (added by #260, PR #261) contains a container, a
`FluentUIFieldComponentBase<TModel, TValue>`, seven `FieldRendererBase` subclasses and 68 tests. Its
`README.md` carries an explicit **"Not yet covered"** section, which is the checklist this issue
turns into work.

What makes the remaining work *cheaper than it was in #260*: **#203 landed on `dev`** (PR #250,
"render collection item fields through `IFieldRendererService`"). When #260 was planned, adding
collection support to a new adapter meant reimplementing `CollectionFieldComponent`'s imperative
`RenderTreeBuilder` path — code whose planned fate was deletion. That is no longer true. Item fields
now resolve through the same `IFieldRenderer` registry as ordinary fields, so a Fluent adapter that
registers renderers gets item fields for the field types it already supports, close to free.

What makes part of it *more urgent than a feature gap*: the security guard. The Fluent container's
`OnInitializedAsync` throws when `Configuration.Security is not null`. That was the right call for
#260 — a security feature that silently does nothing is worse than one that is absent — but it means
`.WithSecurity(...)` and Fluent are mutually exclusive today.

### Approaches

**A. One parity PR.** Everything above in a single change. Rejected on reviewability: #261 was 50
files for five field types; this is a larger surface plus the security pipeline plus the demo app.

**B. Priority-ordered increments, security first.** Six independently shippable slices in the order
above, each its own PR, each leaving the package coherent. Costs more PR overhead; every intermediate
state is releasable and the most-blocking item is fixed first. **Recommended.**

**C. Field types first, security last.** The path of least resistance — the field renderers are
mechanical, the security pipeline is not. Rejected: it optimises for the implementer's comfort over
the user's. A Fluent user blocked on `.WithSecurity(...)` stays blocked through five other PRs.

### Recommendation

**Approach B.** The ordering is the whole point. Two notes carried into the spec:

- **Security enforcement should reuse, not re-copy.** `FormCraftComponent.razor.cs` in the MudBlazor
  package holds ~150 lines of `EnforceSecurityAsync` / `LogSubmissionAuditEventAsync` /
  `GetEncryptedFieldValues` that reference no MudBlazor type. Copying them into the Fluent container
  would create the same duplication the sibling issue exists to remove — so that issue is a
  **soft dependency** of step 1 here, and doing it first makes this cheaper.
- **Item-field support should be verified by a parity test**, in the spirit of
  `RenderPipelineParityTests`: the same configuration rendered inside and outside `.WithItemForm(...)`
  must produce the same presentation attributes. That suite is what has repeatedly caught this exact
  class of drift on the MudBlazor side (#146, #177, #184, #190).

</details>

<details>

### Goal

`FormCraft.ForFluentUI` supports every capability `FormCraft.ForMudBlazor` does, delivered as six
independently shippable increments in priority order.

### Scope

Six slices, each its own PR:

| # | Slice | Removes |
|---|---|---|
| 1 | `.WithSecurity(...)` enforcement | the `NotSupportedException` guard |
| 2 | Collection / item-form fields | "collection fields not covered" |
| 3 | Field-group layout (card, columns, collapsible) | flat-list rendering |
| 4 | Lookup, LOV, autocomplete, multi-select, file upload | five "not covered" entries |
| 5 | Custom renderers + Fluent demo page | the last rendering gaps |
| 6 | Typed `.WithNativeRequired(...)` for Fluent | the raw-attribute workaround |

**Non-goals:** changing `FormCraft` core's public API; changing `FormCraft.ForMudBlazor`'s behaviour;
the shared-machinery move (its own issue).

### Where each slice lands

```mermaid
flowchart TD
    A["FormCraftComponent&lt;TModel&gt;<br/>(Fluent container)"] --> B["slice 1: EnforceSecurityAsync<br/>replaces the throw"]
    A --> C["slice 3: field-group layout<br/>FluentCard + FluentGrid"]
    A --> D["IFieldRendererService"]
    D --> E["slice 2: item fields<br/>(free since #203 — same registry)"]
    D --> F["slice 4: lookup / LOV / autocomplete<br/>multi-select / file upload renderers"]
    D --> G["slice 5: custom renderers"]
    H["slice 6: WithNativeRequired&lt;TModel,TValue&gt;"] -.depends on.-> I["shared-machinery move<br/>(sibling issue)"]
    B -.reuses.-> I

    style B fill:#ffd7d7,stroke:#b02a2a
    style E fill:#cfe6ff,stroke:#2b6cb0
```

### Behaviour rules

1. **Security parity is behavioural, not structural.** Slice 1 is done when a Fluent form with
   `.WithSecurity(s => s.WithRateLimit(...).EnableCsrfProtection())` blocks an over-limit submission
   and rejects an invalid token — asserted by tests mirroring
   `FormCraftComponentSecurityTests`, not by the guard merely being deleted.
2. **Item fields must agree with ordinary fields.** A Fluent parity test asserts that the same field
   configuration renders the same presentation attributes inside and outside `.WithItemForm(...)`,
   including `aria-required`.
3. **`aria-required` stays adapter-owned.** #260 measured that Fluent v5 does **not** emit it from its
   `Required` parameter. Every new field type must write it explicitly via
   `FluentUIFieldComponentBase.AriaRequired`, and every new field type gets an `aria-required` test.
4. **Help text stays container-rendered.** #260 measured that `FluentTextInput`/`FluentNumberInput`
   overwrite a caller-supplied `Message`. New field types use the container's `FieldHelpText` +
   `aria-describedby`, not the `Message` parameter.
5. **The README's "Not yet covered" list shrinks with each slice** — it is the user-visible contract.

### Assumptions

- Fluent UI v5 provides usable equivalents for every remaining type (`FluentAutocomplete<,>`,
  `FluentInputFile`, `FluentSlider<T>`, `FluentColorPicker`, `FluentRatingDisplay`, `FluentDialog` +
  `FluentDataGrid` for lookup/LOV). All exist in `5.0.0-rc.5`; their APIs are unverified.
- Slice 1 lands after the shared-machinery move, or copies the security members as an accepted
  interim duplication if that issue has not landed.

</details>

## 🛠️ Implementation plan

**Goal:** Bring `FormCraft.ForFluentUI` to full parity with `FormCraft.ForMudBlazor`, in six
independently shippable slices ordered by how much each currently blocks users.

**Architecture:** Additive work inside `FormCraft.ForFluentUI` only — new `FieldRendererBase`
subclasses registered in `AddFormCraftFluentUI()`, new components deriving from
`FluentUIFieldComponentBase<TModel, TValue>`, plus container changes for security and field groups.
No change to `FormCraft` core or to the MudBlazor adapter.

**Tech Stack:** .NET 8 / .NET 10 (`Microsoft.NET.Sdk.Razor`), Fluent UI Blazor 5.x, xUnit v3 + bUnit +
Shouldly, NUKE, MinVer.

### Global Constraints

- Target frameworks `net8.0;net10.0`; `TreatWarningsAsErrors=true` — never relaxed to make a build pass.
- Central package management: versions in `Directory.Packages.props`, no `Version` in `.csproj`.
- No `<Version>` element anywhere — MinVer derives it from git tags. Never hand-edit `CHANGELOG.md`.
- `dotnet test --filter` is inert under MTP; verify with the full run `dotnet test -c Release`.
- `FormCraft` core stays UI-framework-agnostic — no Fluent type may leak into it.
- Every new field type writes `aria-required` explicitly and gets a test for it (#199, #260).
- Every new field type takes help text from the container's `FieldHelpText`, not Fluent's `Message`.
- Commit identity: `git -c user.email=phmatray@gmail.com -c user.name="Philippe Matray"`.
- Conventional Commits, scope `fluentui`.

---

### Task 1: Enforce `.WithSecurity(...)` and remove the throw

**Files:** modify `FormCraft.ForFluentUI/Features/FormContainer/FormCraftComponent.razor.cs` and
`.razor`; test `FormCraft.ForFluentUI.UnitTests/Components/FormCraftComponentSecurityTests.cs`;
delete `FormCraft.ForFluentUI.UnitTests/Components/UnsupportedSecurityTests.cs`.

**Interfaces:** container gains `EnforceSecurityAsync()`, `LogSubmissionAuditEventAsync(string, string?)`,
`GetEncryptedFieldValues()` returning `IReadOnlyDictionary<string, string?>`, and a `_securityError`
banner rendered via `FluentMessageBar`.

- [x] **Step 1:** Write the failing test — a form with `.WithSecurity(s => s.WithRateLimit(1, TimeSpan.FromMinutes(1)))` renders, and a second submission is blocked with a visible error. Model it on `FormCraft.ForMudBlazor.UnitTests/Components/FormCraftComponentSecurityTests.cs`.
- [x] **Step 2:** Run `dotnet test -c Release` → FAIL (the container still throws `NotSupportedException`).
- [x] **Step 3:** Port the security members from the MudBlazor container — prefer calling the shared implementation if the sibling machinery-move issue has landed; otherwise copy them and note the duplication in an XML comment.
- [x] **Step 4:** Render the security-error banner with `FluentMessageBar` (`Intent="MessageIntent.Error"`), carrying `data-testid="formcraft-security-error"` to match the MudBlazor test hook.
- [x] **Step 5:** Delete the `NotSupportedException` guard and `UnsupportedSecurityTests.cs`, which asserted the behaviour being replaced.
- [x] **Step 6:** Add CSRF and audit-logging tests mirroring the MudBlazor suite.
- [x] **Step 7:** Run `dotnet test -c Release` → PASS.
- [x] **Step 8:** Update `FormCraft.ForFluentUI/README.md` and the root `README.md` to drop `.WithSecurity(...)` from "Not yet covered".
- [x] **Step 9:** Commit: `feat(fluentui): enforce WithSecurity settings in the Fluent container`.

---

### Task 2: Collection / item-form fields

**Files:** modify `FormCraft.ForFluentUI/Extensions/ServiceCollectionExtensions.cs`; create
`FormCraft.ForFluentUI/Features/CollectionField/` components as needed; test
`FormCraft.ForFluentUI.UnitTests/Fields/CollectionFieldTests.cs` and
`FormCraft.ForFluentUI.UnitTests/Components/RenderPipelineParityTests.cs`.

**Interfaces:** a Fluent collection renderer registered in `AddFormCraftFluentUI()`, ahead of the
type-based renderers.

- [x] **Step 1:** Read `FormCraft.ForMudBlazor/Features/CollectionField/` as it stands **after #203** — item fields now dispatch through `IFieldRendererService`, so establish how much is genuinely MudBlazor-specific before writing anything.
- [x] **Step 2:** Write the failing test — a `.WithItemForm(...)` collection renders one Fluent text input per item and an add/remove control.
- [x] **Step 3:** Run `dotnet test -c Release` → FAIL.
- [x] **Step 4:** Implement the Fluent collection component and renderer, dispatching item fields through `IFieldRendererService` exactly as the post-#203 MudBlazor path does.
- [x] **Step 6:** Add `RenderPipelineParityTests` for Fluent — the same configuration inside and outside `.WithItemForm(...)` must produce the same `aria-required`, label, placeholder and disabled state.
- [x] **Step 7:** Run `dotnet test -c Release` → PASS.
- [x] **Step 8:** Commit: `feat(fluentui): render collection item fields through the renderer registry`.

---

### Task 3: Field-group layout

**Files:** modify `FormCraft.ForFluentUI/Features/FormContainer/FormCraftComponent.razor` and
`.razor.cs`; test `FormCraft.ForFluentUI.UnitTests/Components/FieldGroupTests.cs`.

**Interfaces:** container renders `IGroupedFormConfiguration<TModel>` groups using `FluentCard` and
`FluentGrid`/`FluentGridItem`, honouring `Columns`, `ShowCard` and `Name`.

- [x] **Step 1:** Write the failing test — a form with two `.AddFieldGroup(...)` groups renders two cards, each showing its group name.
- [x] **Step 2:** Run `dotnet test -c Release` → FAIL (fields render as a flat list).
- [x] **Step 3:** Implement grouped rendering in the container, modelled on the MudBlazor container's `RenderFieldGroup`/`RenderUngroupedFields` but using `FluentCard` + `FluentGrid`.
- [x] **Step 4:** Run `dotnet test -c Release` → PASS.
- [x] **Step 5:** Add a test that ungrouped fields still render when a grouped configuration is present.
- [x] **Step 6:** Run `dotnet test -c Release` → PASS.
- [x] **Step 7:** Commit: `feat(fluentui): render field groups with card and column layout`.

---

### Task 4: Lookup, LOV, autocomplete, multi-select and file upload

**Files:** create `FormCraft.ForFluentUI/Fields/{LookupField,LovField,AutocompleteField,SelectField,FileUploadField}/`
components and renderers; modify `FormCraft.ForFluentUI/Extensions/ServiceCollectionExtensions.cs`;
tests per field type under `FormCraft.ForFluentUI.UnitTests/Fields/`.

**Interfaces:** `FluentUILookupFieldRenderer`, `FluentUILovFieldRenderer`,
`FluentUIAutocompleteFieldRenderer`, `FluentUIMultiSelectFieldRenderer`,
`FluentUIFileUploadFieldRenderer`, `FluentUIMultipleFileUploadRenderer` — all `FieldRendererBase`,
all registered **above** the type-based block.

- [x] **Step 1:** Confirm the v5 component APIs from the package itself (`FluentAutocomplete<,>`, `FluentInputFile`, `FluentDialog`, `FluentDataGrid<T>`) rather than from documentation — #260 found several v4→v5 renames this way.
- [x] **Step 2:** Write the failing test for autocomplete — a field configured with an async search renders `FluentAutocomplete` and returns matches.
- [x] **Step 3:** Run `dotnet test -c Release` → FAIL.
- [x] **Step 4:** Implement the autocomplete component + renderer; register it above the type-based renderers.
- [x] **Step 5:** Run `dotnet test -c Release` → PASS. Commit: `feat(fluentui): render autocomplete fields`.
- [x] **Step 6:** Repeat steps 2-5 for multi-select (`FluentSelect` with `Multiple`), asserting the selected set round-trips to the model.
- [x] **Step 7:** Repeat steps 2-5 for file upload, single and multiple (`FluentInputFile`).
- [x] **Step 8:** Repeat steps 2-5 for lookup (`FluentDialog` + `FluentDataGrid<T>`).
- [x] **Step 9:** Repeat steps 2-5 for LOV.
- [x] **Step 10:** Add an `aria-required` test per new field type, and update both READMEs' "Not yet covered" lists.
- [x] **Step 11:** Run `dotnet test -c Release` → PASS. Commit: `feat(fluentui): render lookup and LOV dialog fields`.

---

### Task 5: Custom renderers and the demo page

**Files:** create `FormCraft.ForFluentUI/Features/CustomFields/` (colour picker, rating, slider);
modify `FormCraft.DemoBlazorApp` to add a Fluent showcase page; tests under
`FormCraft.ForFluentUI.UnitTests/Fields/`.

**Interfaces:** `FluentUIColorPickerRenderer`, `FluentUIRatingRenderer`, `FluentUISliderRenderer` —
custom renderers used via `.WithCustomRenderer(...)`, not registered as `IFieldRenderer`.

- [x] **Step 1:** Write the failing test — a field with `.WithCustomRenderer(new FluentUISliderRenderer())` renders `FluentSlider<T>`.
- [x] **Step 2:** Run `dotnet test -c Release` → FAIL.
- [x] **Step 3:** Implement the three custom renderers, mirroring `FormCraft.ForMudBlazor/Features/CustomFields/`.
- [x] **Step 4:** Run `dotnet test -c Release` → PASS.
- [x] **Step 5:** Add a Fluent showcase page to `FormCraft.DemoBlazorApp`. ⚠️ The demo registers `AddFormCraftMudBlazor()`; the two adapters are mutually exclusive in one container, so the page needs its own scoped provider or the demo needs splitting — decide and record the choice in the PR.
- [x] **Step 6:** Run `dotnet build -c Release && dotnet test -c Release` → PASS.
- [x] **Step 7:** Commit: `feat(fluentui): add custom renderers and a demo showcase page`.

---

### Task 6: Typed `.WithNativeRequired(...)` for Fluent

**Files:** create `FormCraft.ForFluentUI/Extensions/FieldBuilderExtensions.cs` (or consume the shared
one, if the machinery-move issue landed); test
`FormCraft.ForFluentUI.UnitTests/Extensions/NativeRequiredBuilderTests.cs`.

**Interfaces:** `WithNativeRequired<TModel, TValue>(this FieldBuilder<TModel, TValue>, bool enabled = true)`
writing the `"Required"` additional attribute.

- [x] **Step 1:** Check whether the shared-machinery issue has moved `WithNativeRequired` into core. If it has, this task is only a test that Fluent honours it — skip to step 4.
- [x] **Step 2:** Write the failing test — `.WithNativeRequired(false)` on a `.Required(...)` Fluent field suppresses `aria-required`, and `.WithNativeRequired()` alone adds it.
- [x] **Step 3:** Implement the extension in a **distinct static class and namespace** from the MudBlazor one. ⚠️ Same name in namespace `FormCraft` would be `CS0121`-ambiguous for any project referencing both packages — which the Fluent test project does.
- [x] **Step 4:** Run `dotnet test -c Release` → PASS.
- [x] **Step 5:** Update `FormCraft.ForFluentUI/README.md` to document the typed opt-out in place of the raw `.WithAttribute("Required", false)` form.
- [x] **Step 6:** Commit: `feat(fluentui): add a typed WithNativeRequired builder extension`.

phmatray:
Implements #278.

Closes #278.

code-review pass.

> **Note on shape.** The issue's spec envisioned these six slices as six separate PRs. They land here
> as **one PR with one commit per slice**, which is what `implement-issue` executes for a single
> issue with a single plan; the commits are separable if a reviewer prefers to split them.

### Plan
- [x] Task 1: Enforce `.WithSecurity(...)` and remove the throw
- [x] Task 2: Collection / item-form fields
- [x] Task 3: Field-group layout
- [x] Task 4: Lookup, LOV, autocomplete, multi-select and file upload
- [x] Task 5: Custom renderers and the demo page
- [x] Task 6: Typed `.WithNativeRequired(...)` for Fluent

## Decisions worth a reviewer's eye

Three places where the plan's assumed Fluent v5 API did not survive contact with the package, each
measured rather than assumed (the discipline #260 established):

1. **Lookup and LOV pickers are inline panels, not modals.** v5's `FluentDialog` is service-driven
   and renders nothing unless the host app places a `FluentDialogProvider` in its layout — which a
   field component cannot verify. A Browse button that silently did nothing on an app without the
   provider is the failure class this repo refuses elsewhere, so the picker renders where it cannot
   fail to appear.
2. **The rating renderer is a row of labelled buttons, not `FluentRatingDisplay`.** That component
   exposes `Value` but no `ValueChanged` and no click handling — it is genuinely read-only, so a
   field bound to it would display a score and silently refuse every edit.
3. **The Fluent showcase is a second app (`FormCraft.DemoFluentApp`), not a page in the existing
   demo.** The adapters are mutually exclusive in one DI container and Blazor has no per-subtree
   service provider, so no amount of injection makes a Fluent page work inside the MudBlazor demo.
   ⚠️ This adds a second Blazor WASM project to the solution, so CI builds one more WASM app.

Also worth noting: `.AsLookup(...)` previously existed **only** in the MudBlazor package, so a
Fluent-only app could not configure a lookup field at all. This PR adds a Fluent one in
`FormCraft.ForFluentUI.Extensions` (a distinct namespace, per the `CS0121` hazard the plan flags for
Task 6). Both write the same attribute keys, so either renders under either adapter.

## Merged with `dev` after #279 landed

`dev` moved six commits ahead mid-flight, including
**`refactor(core)!: move the UI-agnostic adapter machinery into core (#279)`** — the shared-machinery
change this plan named as a soft dependency. The merge had one textual conflict and two *semantic*
ones a clean text merge would have hidden:

- **`ValidateCollections` flipped to `true`.** #279 added it as `false` with the note "turn it on
  when this container actually renders collection fields — not before". Task 2 is that moment, so
  leaving it `false` would have accepted submissions violating collection validators, silently.
  `CollectionFieldSubmitTests` — written by #279 around the old premise — is rewritten to assert the
  new behaviour, including that a blocked submit now has a visible add control the user can act on.
- **Task 6's `WithNativeRequired` is gone from this package.** #279 put it in core, which is the
  outcome the plan's Task 6 Step 1 anticipated. Keeping a Fluent copy would have been worse than
  redundant: core's lives in namespace `FormCraft`, which every consumer imports for `FormBuilder`,
  so two identical signatures would make every call `CS0121`-ambiguous. What remains is the test
  that Fluent honours core's method, via `NativeRequired.Resolve`.

Re-verified on the merged tree: build clean at 0 warnings, 1,524 tests green.

## Follow-ups

Not done here, deliberately — each is adjacent work rather than this issue's scope:

- **Move the security pipeline into shared machinery.** `EnforceSecurityAsync` /
  `LogSubmissionAuditEventAsync` / `GetEncryptedFieldValues` are still a line-for-line copy in both
  containers, referencing no UI type; the copy is marked with a comment saying so. #279 moved the
  *other* UI-agnostic machinery into core but not this, so the duplication remains and is worth its
  own pass.
- **`.WithCustomRenderer<TRenderer>()` is unreachable.** The generic extension is shadowed by the
  instance method `WithCustomRenderer(IFieldRenderer)`, so callers must use the `typeof(...)`
  overload. Pre-existing in core, not introduced here.
- **`AddFormCraftFluentUI()`'s conflict guard is still one-directional** — registering MudBlazor
  *after* Fluent is not caught. Carried over from #260.
- **Unrelated to this PR, surfaced incidentally:** a code-review pass that mis-targeted its range
  ended up reviewing `1ea5425` (the #266 mask work already on `dev`) and reported that
  `.WithMask(() => new BlockMask(...))` binds no mask at all — `TextMaskMap.Resolve` rejects a
  factory-produced mask on `string.IsNullOrWhiteSpace(produced?.Mask)`, and `BlockMask.Mask` is
  `null` even though the mask works. It claims to have measured this on MudBlazor 9.8.0. **Not
  verified here and not touched by this branch** — worth its own issue if it reproduces, since
  `WithMask`'s own XML doc advertises `BlockMask`.


phmatray:
**Follow-up triage after merge.**

One item from this PR's `## Follow-ups` list is **already fixed** and was not filed:

- ~`AddFormCraftFluentUI()`'s conflict guard is one-directional~ — resolved by #279, which landed on `dev` while this PR was in flight. Both adapters now call `AdapterRegistration.EnsureSingleAdapter` in core, so either registration order throws identically. The bullet was written before that merge and was stale by the time this landed.

The remaining three were verified and filed as issues; see the linked issues above/below.

Also worth recording, since it cost a cycle: the first `code-review` pass on this PR ran in the wrong checkout, reported `dev...HEAD` as empty, and silently fell back to reviewing an unrelated already-merged commit. Its findings were about `MaskedValueDiagnostic`, not this branch. Re-running the review against the PR *number* rather than a cwd-relative range produced the real 11 findings.

PR #293, opened 2026-08-13
fix(mudblazor): report post-init and partially-discarded masked values (#283)

### LINKED ISSUE #283: The masked-value diagnostic under-reports: post-init values and partial discards go unwarned
## Problem / motivation

Follow-up from #266 (PR #274). `MaskedValueDiagnostic` warns when a configured mask renders a stored
value as blank. Its coverage boundary has **two known gaps**, both surfaced by #274's code review and
deliberately deferred there because widening either is a spec decision rather than a defect fix.

**(a) It only fires from `OnInitialized`, so a model populated *after* first render never warns.**

```csharp
// The field renders before the fetch resolves. CurrentValue is "" at OnInitialized,
// the diagnostic returns early, and nothing ever warns.
protected override async Task OnInitializedAsync()
    => _model.Phone = await _api.GetPhoneAsync();   // "N/A"
```

`FieldComponentBase.OnParametersSet` later calls `LoadValueFromModel()`, the field renders blank, and
the model keeps `"N/A"` — the exact bug #266 exists to report, unreported. The same holds for a
`.DependsOn(...).WithValueProvider(...)` that computes a non-conforming value, and for a form reset.
This is arguably *the* canonical legacy-data case: data loaded from a store asynchronously.

Worth noting the spec's stated reason for the "initial value" framing — *"a field whose value
legitimately becomes blank (the user cleared it) must not warn on a later render"* — is already
covered independently, because `Applies` requires a **non-blank** configured value. Clearing a field
can never satisfy the rule. So the stated obstacle to emitting later does not actually bite.

**(b) `Applies` requires *total* collapse, so a mask that discards *part* of a value is silent.**

| Stored value | Mask | Renders | Warns today |
|---|---|---|---|
| `"N/A"` | `(000) 000-0000` | `""` | ✅ yes |
| `"+1 555 123 4567"` | `(000) 000-0000` | `(155) 512-3456` | ❌ **no** |
| `"N/A5551234567"` | `(000) 000-0000` | `(555) 123-4567` | ❌ **no** |
| `"5551234567"` | `(000) 000-0000` | `(555) 123-4567` | ✅ no (correct — reformat) |

Row 2 is **a different phone number** than the one stored, displayed to the user, while the model
keeps the original. That is the same display-vs-model divergence the diagnostic was written for, and
it is arguably worse than the blank case: a blank field is visibly wrong, a plausible-but-wrong phone
number is not.

## Proposed solution

Treat the two gaps as one question — *what is the honest boundary of "the mask silently changed what
the user's data means"?* — and widen the rule and the emit point together.

- **(a)** Move the emit off `OnInitialized`-only and onto the external-model-change signal the base
  class already has. `ShouldReloadValue()` distinguishes an external model change from an in-flight
  user edit, so it can drive a re-check without warning on every keystroke. The existing per-field
  latch caps the result at one warning regardless.
- **(b)** Compare the mask's **clean text** against the stored value's mask-relevant characters rather
  than testing for total collapse. `PatternMask.GetCleanText()` gives what survived; if the surviving
  characters are a strict subset of what was stored, input was discarded. A plain reformat keeps every
  significant character and only adds literals, so it still stays silent.

## Alternatives considered

**Warn whenever masked ≠ stored.** Rejected, and it is the tempting one — it is what #266 explicitly
ruled out. It fires on every correctly-masked field in every form (`5551234567` → `(555) 123-4567` is
the *happy path*), and a diagnostic that fires on the happy path is one developers mute, taking the
real signal with it.

**Leave (b) alone and fix only (a).** Reasonable, and a valid way to split this issue if (b) proves
contentious. (a) is a straightforward coverage bug; (b) is a genuine design question.

**Write the masked value back to the model.** Rejected here for the same reason #266 rejected it: it
resolves the divergence by destroying the stored value.

## Area

`FormCraft.ForMudBlazor` — `Diagnostics/MaskedValueDiagnostic.cs`, `Fields/TextField/MudBlazorTextFieldComponent.razor.cs`, and `Fields/MudBlazorFieldComponentBase.cs`.

**Related:** #266, #274, #211, #265

<details>

### Problem / context

#266 shipped a deliberately narrow rule, and the narrowness was correct as a *starting* point: it
guaranteed no false positives on the overwhelmingly common correctly-masked field. #274's review then
probed the boundary and found two classes of true negative. Neither is a mistake in #266 — both are
the cost of the narrow rule, now measured.

The two gaps are independent in mechanism but identical in consequence: **the user sees something
other than what the model holds, submits, and the original survives.** That shared consequence is why
they belong in one issue rather than two.

### Approaches

**A. Fix both together, framed as "what counts as the mask changing the data's meaning".** Emit on
external model change as well as init, and replace the collapse test with a discard test. One
coherent rule, one pass over the tests. Risk: (b) can produce false positives if the discard test is
wrong, and bundling it with (a) means a rollback loses the uncontroversial half.

**B. Fix (a) only; close (b) as working-as-intended.** Cheapest and safest. Leaves a real divergence
class unreported — the plausible-but-wrong-value case, which is the harder one for a developer to
notice unaided.

**C. Fix (a), and put (b) behind an opt-in strictness setting.** Both behaviours available, nobody
surprised. Costs a public configuration surface for a diagnostic, which is a lot of API for a warning,
and splits the "what is correct?" question into "what do you want?" — usually a sign the rule has not
been decided.

### Recommendation

**A**, with the tasks ordered so (a) lands first and independently. (a) is a coverage bug with no
design risk; (b) is the design change. Sequencing them in that order means the uncontroversial fix is
already committed and revertible on its own if (b) needs rethinking. Reject **C** — a strictness knob
is what you add when you cannot decide the rule, and this rule is decidable.

</details>

<details>

### Goal

A masked field whose stored value the mask *changes the meaning of* — whether by blanking it or by
discarding part of it — is reported once, whenever that value arrives, not only if it happened to be
present at first render.

### Scope

- Emit the diagnostic on external model change, not only from `OnInitialized`.
- Replace the total-collapse rule with a discard rule that still ignores pure reformatting.
- Tests for both, including the negatives that keep the rule honest.

### Non-goals

- Changing what renders. The field still shows what the mask produces.
- Writing the masked value back to the model, ever.
- Warning on a plain reformat (`5551234567` → `(555) 123-4567`).
- Warning on a value the *user* typed — this reports stored data, not live editing.

### Behaviour

| Stored value | Mask | Masked result | Warns |
|---|---|---|---|
| `"N/A"` | `(000) 000-0000` | `""` | yes (unchanged from #266) |
| `"+1 555 123 4567"` | `(000) 000-0000` | `(155) 512-3456` | **yes (new)** |
| `"5551234567"` | `(000) 000-0000` | `(555) 123-4567` | no — reformat |
| `"N/A"` arriving after first render | `(000) 000-0000` | `""` | **yes (new)** |
| user clears the field | any | `""` | no — value is blank |
| same field, 50 collection rows | any | — | once |

### Key files

- modify `FormCraft.ForMudBlazor/Diagnostics/MaskedValueDiagnostic.cs` — the rule.
- modify `FormCraft.ForMudBlazor/Fields/TextField/MudBlazorTextFieldComponent.razor.cs` — the emit point.
- tests: `FormCraft.ForMudBlazor.UnitTests/Diagnostics/MaskedValueDiagnosticTests.cs`, `Fields/MudBlazorTextFieldComponentTests.cs`, `Fields/CollectionInputTypeTests.cs`.

### Validation rules

- Fires when the configured value is non-blank **and** the mask either blanked it or discarded characters from it.
- Never fires on a value the mask only reformatted.
- Fires at most once per field per component lifetime, latched on the existing category-qualified key.
- Never fires when no mask resolves — including the `MaskFactory` path added by #265.

### Edge cases

- **The latch must not be consumed by a non-warning check.** #274 established this: `ShouldWarnOnce` has a side effect, so it must be consulted only after the rule says yes, or row order decides whether a collection field is ever reported.
- **Re-checking on model change must not re-check on every keystroke.** `ShouldReloadValue()` is the signal that distinguishes them; the latch is the backstop.
- **`CleanDelimiters` (#265) changes what "surviving characters" means** — a mask that strips its own literals must not read as having discarded them.

### Assumptions

- `PatternMask.GetCleanText()` returns the mask-significant characters; verify against MudBlazor 9.8.0 before relying on it, and fall back to comparing against the mask's own literal set if not.
- The diagnostic stays `internal`, tested via the `InternalsVisibleTo` added in #274.

</details>

## 🛠️ Implementation plan

**Goal:** report a mask that blanks *or* partially discards a stored value, whenever that value arrives.
**Architecture:** the rule stays in the diagnostic type; the emit point moves to the base class's external-reload signal.

**Global Constraints**

- **Both render paths move together** — item fields route through the same component since #203, so one emit site covers both; do not add a second.
- **Never mutate the model.** This reports; it does not repair.
- **Consult the latch only after the rule says yes** — `ShouldWarnOnce` has a side effect (#274).
- **Resolve the mask via `GetMask()`**, never `TextMaskMap.Resolve(Mask)` — #265 means a mask can come from a factory with no pattern string.
- **`dotnet test --filter` is inert here** (`MTP0001`) — run `dotnet test -c Release`. (See #277.)

### Task 1: Emit when the value arrives after first render

**Files:** modify `FormCraft.ForMudBlazor/Fields/TextField/MudBlazorTextFieldComponent.razor.cs`; test `FormCraft.ForMudBlazor.UnitTests/Fields/MudBlazorTextFieldComponentTests.cs`.

**Interfaces:** call `WarnIfMaskBlanksTheStoredValue()` from the external-model-change path as well as `OnInitialized`, gated on `ShouldReloadValue()`.

- [x] **Step 1:** Write a failing test: render a masked field over an empty model, then set the model's value to a non-conforming one and re-render — expect exactly one warning.
- [x] **Step 2:** Write a second failing-safe test: typing into the field (a user edit, not an external change) produces no warning.
- [x] **Step 3:** Run `dotnet test -c Release` → FAIL (nothing logged for the async case).
- [x] **Step 4:** Move the call so it also runs on external model change, reusing the existing latch; confirm `ShouldReloadValue()` is the gate that excludes in-flight user edits.
- [x] **Step 5:** Run `dotnet test -c Release` → PASS, and confirm the existing #266 tests still pass unchanged.
- [x] **Step 6:** Commit: `fix(mudblazor): report a mask-blanked value that arrives after first render`

### Task 2: Verify what MudBlazor actually gives us for a partial discard

**Files:** test `FormCraft.ForMudBlazor.UnitTests/Diagnostics/MaskedValueDiagnosticTests.cs`.

**Interfaces:** none yet — this task establishes ground truth before the rule is written.

- [x] **Step 1:** Write characterisation tests recording what `PatternMask("(000) 000-0000")` produces for `"+1 555 123 4567"`, `"N/A5551234567"`, `"5551234567"` and `"N/A"` — both `Text` and `GetCleanText()`.
- [x] **Step 2:** Run `dotnet test -c Release` → they pass by construction; read the recorded values.
- [x] **Step 3:** Confirm from those values that a discard is distinguishable from a reformat; if `GetCleanText()` cannot express the difference, note the fallback (compare against the mask's literal set) in the test's comment before proceeding.
- [x] **Step 4:** Commit: `test(mudblazor): characterise PatternMask output for partial discards`

### Task 3: Widen the rule to catch a partial discard

**Files:** modify `FormCraft.ForMudBlazor/Diagnostics/MaskedValueDiagnostic.cs`; test `FormCraft.ForMudBlazor.UnitTests/Diagnostics/MaskedValueDiagnosticTests.cs`.

**Interfaces:** `Applies` gains whatever input Task 2 proved is needed (likely the clean text alongside the rendered text); update the single call site.

- [x] **Step 1:** Write failing tests for the new rows: `"+1 555 123 4567"` → warns; `"N/A5551234567"` → warns; `"5551234567"` → still silent; `"N/A"` → still warns.
- [x] **Step 2:** Run `dotnet test -c Release` → FAIL (discard cases silent).
- [x] **Step 3:** Implement the discard rule, keeping the blank case as a subset of it so no #266 behaviour regresses.
- [x] **Step 4:** Add a `CleanDelimiters: true` case (#265) proving a mask that strips its own literals does not read as a discard.
- [x] **Step 6:** Commit: `feat(mudblazor): report a mask that discards part of a stored value`

### Task 4: Update the README

**Files:** modify `README.md`.

- [x] **Step 1:** Update the `Mask` Unreleased entry's point 2 — it currently says the diagnostic fires only on outright rejection; state the widened rule and that a post-init value is covered.
- [x] **Step 2:** Run `dotnet build -c Release` and `dotnet test -c Release` → both green.
- [x] **Step 3:** Commit: `docs: document the widened masked-value diagnostic`

phmatray:
Implements #283.

Closes #283.

Widens `MaskedValueDiagnostic` on two axes deferred from #274:

- **(a)** it fires on external model change, not only from `OnInitialized` — so a value fetched after first render is reported;
- **(b)** it reports a mask that *discards part of* a stored value, not only one that blanks it outright, while staying silent on a plain reformat.

### Plan
- [x] Task 1: Emit when the value arrives after first render
- [x] Task 2: Verify what MudBlazor actually gives us for a partial discard
- [x] Task 3: Widen the rule to catch a partial discard
- [x] Task 4: Update the README

### Code review

A review of `dev...HEAD` found — and reproduced end-to-end — a false-positive class in the first draft of the rule, plus several smaller issues. All fixed in `b2bf82f`:

- **The rule stripped only the *mask's* literals from both sides**, so a stored value punctuated any other way was reported as a discard with nothing lost: `000-00-0000` + `"123 45 6789"`, `000.000.0000` + `"555-123-4567"`, and a credit card with dashes under a space-separated mask all warned. The whole suite was blind to it because every mask test used `(000) 000-0000`, whose decoration `"() -"` happens to contain the separators the tests supplied — so the row described as "the one that keeps it honest" passed for the wrong reason. The rule now keeps a character only if it could match a placeholder **and** is not decoration the mask contributes; stripping non-alphanumerics alone was rejected because a pattern can spell an alphanumeric literal (`+1 000-0000`). Pinned by cases that pair a pattern with a value punctuated differently, at both the rule and component level — verified to fail 9 tests against the old rule.
- **`PatternMask.Placeholder` was not treated as decoration**, so a mask that pads short values (`(555) 123-45__`) warned on essentially every value of a variable-length field.
- **The instance latch was consulted after the mask had already been resolved and run**, so the now-hot `OnParametersSet` path re-did the work — including invoking a caller-supplied mask factory — on every external model change forever. Hoisted, and now also set when a sibling row has already claimed the shared latch.
- Pattern and decoration are read **after** `SetText`, since `MultiMask` rewrites its own `Mask` as it matches.
- `Warn` no longer re-derives which case applies; both it and `Applies` call one `RendersEmpty` predicate, so the rule and the message cannot drift.
- Renamed `WarnIfMaskBlanksTheStoredValue` → `WarnIfMaskChangesTheStoredValue` and `LiteralsOf` → `DecorationOf` to match what they now do.

## Follow-ups

Not addressed here, deliberately — both are pre-existing and wider than this issue:

- **Mask attributes are read only in `OnInitialized`.** A component instance re-parameterised with a new `Context` (swapping `Configuration`, or unkeyed collection rows shifting) judges the new value against the previous field's mask. This is pre-existing for *rendering* too — `Mask="@GetMask()"` has the same staleness — so fixing it only for the diagnostic would be incoherent; it wants its own issue covering both.
- **`MultiMask` is untested here.** It derives from `PatternMask` and rewrites its own `Mask` as it matches. Reading the pattern after `SetText` is the right order for it, but nothing pins the behaviour, and I could not get it to switch via `SetText` alone on 9.8.0 — so it is a reasoned assumption rather than a measured one.

### DIFF
--- a/FormCraft.ForMudBlazor.UnitTests/Diagnostics/MaskedValueDiagnosticTests.cs
+++ b/FormCraft.ForMudBlazor.UnitTests/Diagnostics/MaskedValueDiagnosticTests.cs
@@ -77,4 +77,118 @@ public class MaskedValueDiagnosticTests
         // Assert
         applies.ShouldBeFalse();
     }
+
+    // ---------------------------------------------------------------------------------------------
+    // #283 Task 2 — characterisation. Ground truth before the rule, because the rule in Task 3 hangs
+    // entirely on whether MudBlazor gives us something that can tell a DISCARD apart from a
+    // REFORMAT. These tests record what PatternMask 9.8.0 actually does rather than what the issue
+    // assumed it does; if a MudBlazor upgrade changes any of it, these fail first and name the
+    // reason, instead of the rule quietly going wrong.
+    // ---------------------------------------------------------------------------------------------
+
+    /// <summary>
+    /// What <c>PatternMask("(000) 000-0000")</c> produces for the four values in the issue's table.
+    /// </summary>
+    /// <remarks>
+    /// <para>
+    /// <c>Text</c> is what renders. <c>GetCleanText()</c> is what the mask considers the significant
+    /// characters — the value with the pattern's literals stripped — and is what <c>CleanDelimiters</c>
+    /// makes the mask report back to the model (#265).
+    /// </para>
+    /// <para>
+    /// First finding: the mask CONSUMES characters positionally and silently drops whatever does not
+    /// fit. <c>"+1 555 123 4567"</c> keeps only the first ten digits — <i>a different phone
+    /// number</i> — and <c>"N/A5551234567"</c> drops the letters and keeps the digits. Neither
+    /// blanks, so #266's total-collapse rule cannot see either.
+    /// </para>
+    /// <para>
+    /// Second finding, and the one that changed the design: <b><c>GetCleanText()</c> is not "what
+    /// survived".</b> The issue assumed it returns the mask-significant characters; measured against
+    /// 9.8.0 it returns <c>Text</c> verbatim whenever <c>CleanDelimiters</c> is <c>false</c> — which
+    /// is FormCraft's default and therefore the normal case. Pinned by
+    /// <see cref="PatternMask_CleanText_Should_Follow_CleanDelimiters"/>. A rule written on
+    /// <c>GetCleanText()</c> would have compared <c>"(555) 123-4567"</c> against the raw stored value
+    /// and fired on every correctly-masked field — the exact happy-path false positive #266 was
+    /// designed to avoid.
+    /// </para>
+    /// <para>
+    /// So Task 3 takes the fallback the spec named: compare the two sides with the MASK'S OWN
+    /// LITERALS removed from each. <c>"555 123 4567"</c> and <c>"(555) 123-4567"</c> both reduce to
+    /// <c>5551234567</c> and stay silent, while <c>"+1 555 123 4567"</c> reduces to
+    /// <c>+15551234567</c> against the rendered <c>1555123456</c> and warns. The literal set is
+    /// derivable because the mask exposes its placeholder alphabet — see
+    /// <see cref="PatternMask_Should_Expose_Its_Placeholder_Alphabet"/>.
+    /// </para>
+    /// </remarks>
+    [Theory]
+    [InlineData("N/A", "", "")]
+    [InlineData("+1 555 123 4567", "(155) 512-3456", "(155) 512-3456")]
+    [InlineData("N/A5551234567", "(555) 123-4567", "(555) 123-4567")]
+    [InlineData("5551234567", "(555) 123-4567", "(555) 123-4567")]
+    public void PatternMask_Should_Produce_The_Recorded_Text_And_CleanText(
+        string stored,
+        string expectedText,
+        string expectedCleanText)
+    {
+        // Arrange
+        var mask = new PatternMask("(000) 000-0000");
+
+        // Act
+        mask.SetText(stored);
+
+        // Assert
+        mask.Text.ShouldBe(expectedText);
+        mask.GetCleanText().ShouldBe(expectedCleanText);
+    }
+
+    /// <summary>
+    /// <c>GetCleanText()</c> reports the significant characters regardless of <c>CleanDelimiters</c>.
+    /// </summary>
+    /// <remarks>
+    /// The edge case the spec flagged (#265): a mask that strips its own literals must not read as
+    /// having discarded them. This records that <c>CleanDelimiters</c> changes only what the mask
+    /// reports to the MODEL, not what <c>GetCleanText()</c> returns — so a rule written on
+    /// <c>GetCleanText()</c> is unaffected by the setting, and the same stored value yields the same
+    /// verdict either way. Without this, Task 3's rule would need a special case it does not need.
+    /// </remarks>
+    [Theory]
+    [InlineData(false, "(555) 123-4567")]
+    [InlineData(true, "5551234567")]
+    public void PatternMask_CleanText_Should_Follow_CleanDelimiters(
+        bool cleanDelimiters,
+        string expectedCleanText)
+    {
+        // Arrange
+        var mask = new PatternMask("(000) 000-0000") { CleanDelimiters = cleanDelimiters };
+
+        // Act
+        mask.SetText("5551234567");
+
+        // Assert - Text is the same either way; only GetCleanText() moves.
+        mask.Text.ShouldBe("(555) 123-4567");
+        mask.GetCleanText().ShouldBe(expectedCleanText);
+    }
+
+    /// <summary>
+    /// The pattern's placeholder alphabet is readable off the mask, so its literals are derivable.
+    /// </summary>
+    /// <remarks>
+    /// This is the fallback the spec named, and measuring it is what makes the fallback usable rather
+    /// than hypothetical. <c>MaskChars</c> carries the placeholder characters — <c>0</c>, <c>a</c>,
+    /// <c>*</c> by default — so "a literal" is precisely "a pattern character that is not one of
+    /// these", computed from the mask in hand rather than from a hardcoded list that a caller
+    /// supplying custom <c>MaskChars</c> would silently invalidate.
+    /// </remarks>
+    [Fact]
+    public void PatternMask_Should_Expose_Its_Placeholder_Alphabet()
+    {
+        // Arrange
+        var mask = new PatternMask("(000) 000-0000");
+
+        // Act
+        var placeholders = mask.MaskChars.Select(m => m.Char).ToArray();
+
+        // Assert
+        placeholders.ShouldBe(['0', 'a', '*'], ignoreOrder: true);
+    }
 }
--- a/FormCraft.ForMudBlazor.UnitTests/Diagnostics/MaskedValueDiagnosticTests.cs
+++ b/FormCraft.ForMudBlazor.UnitTests/Diagnostics/MaskedValueDiagnosticTests.cs
@@ -1,22 +1,30 @@
 namespace FormCraft.ForMudBlazor.UnitTests.Diagnostics;
 
 /// <summary>
-/// Tests the rule behind the diagnostic that reports a mask blanking a stored value (#266).
+/// Tests the rule behind the diagnostic that reports a mask blanking (#266) or partly discarding
+/// (#283) a stored value.
 /// <para>
-/// The rule is deliberately narrow, and the narrowness is the whole design: a mask that reformats
-/// <c>5551234567</c> into <c>(555) 123-4567</c> is doing its job, and warning about it would fire on
-/// every correctly-masked field in every form until someone muted the category. Only total collapse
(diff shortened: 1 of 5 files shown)


phmatray:
Recording a follow-up observation from this PR rather than filing it, since it is an unverified assumption rather than committed work.

**`MultiMask` behaviour through the mask path is reasoned, not measured.** `MultiMask` derives from `PatternMask` and rewrites its own `Mask` as it matches, which is why `WarnIfMaskChangesTheStoredValue` reads the pattern and the decoration **after** `SetText` rather than before — so both describe the mask that actually produced the rendered text. That ordering is correct by construction, but nothing pins it: on MudBlazor 9.8.0 I could not get a `MultiMask` to switch sub-masks via `SetText` alone, so there is no test that would fail if the order were reversed.

If someone finds a way to drive a sub-mask switch from `SetText`, that is the test to add — `MaskedValueDiagnosticTests` is where it belongs. Until then, treat the read-order comment in `MudBlazorTextFieldComponent.razor.cs` as the record of the assumption.

Issue #205, opened 2026-08-11
Extract a shared collection-item test fixture instead of copying models per suite


phmatray:
## Problem / motivation

The collection-item test suites duplicate their scaffolding wholesale. `CollectionRequiredTests.cs`
(added in #193) and `CollectionAdornmentTests.cs` (#184) both carry their own copies of:

- helpers — `RenderOrderForm(...)`, `BuildConfiguration(Action<FieldBuilder<…>>)`
- models — `OrderModel` / `OrderItem`, `BasketModel` / `BasketLine`, `AppointmentModel` / `AppointmentSlot`

They differ only in incidentals (whether `ProductName` is seeded with `"Widget"`, whether `BasketLine`
declares `IsGift`). Both exercise the **same** `AddCommonFieldAttributes` helper across the **same**
renderers — text, numeric, date, and the boolean path that bypasses it.

This is not a hypothetical tidiness complaint. `RenderPipelineParityTests`' divergence list names the
attribute suites still to be written — the raw `"Required"` key, `InputType` / `Lines` / `MaxLength` /
`Autocomplete` (#189), `OnAdornmentClick`, plus `EnablePasswordToggle` and `Mask`. Each one needs the
same four model pairs and the same two helpers. The next suite is the **third** copy, and the pattern
has no natural stopping point.

There is already a measurable cost. When #193 needed a date-path test, the `AppointmentModel` /
`AppointmentSlot` pair had to be re-declared even though `CollectionAdornmentTests` had defined an
identical pair for exactly the same purpose. And the copied comment in that suite said
`AddCommonFieldAttributes` was "shared by the text and numeric paths" — a miscount that survived the
copy and hid the date path from coverage until review caught it. **Duplicated scaffolding duplicates
its errors too.**

## Proposed solution

Extract a shared collection-item test fixture into the test project — the model pairs and the
render/configure helpers — and have the existing suites use it. New attribute suites then declare only
what they actually test.

The models are the valuable half: one canonical `OrderItem` (string), `BasketLine` (numeric + bool) and
`AppointmentSlot` (date) covering the three renderers fed by the shared helper plus the one that
bypasses it, so every future suite tests all four paths by default instead of by remembering to.

## Alternatives considered

- **Leave it.** Zero risk today, and the duplication is currently only 2×. But the divergence list makes
  the growth predictable, and the miscounted-comment incident shows the copies propagate mistakes as
  well as code.
[...]

PR #294, opened 2026-08-13
refactor(mudblazor): express the two diagnostic idioms once each (#284)

### LINKED ISSUE #284: Diagnostic emission is copy-pasted four times across the MudBlazor package
## Problem / motivation

Follow-up from #274. `FormCraft.ForMudBlazor` now has **four** diagnostics, and emitting one requires
hand-writing two idioms that are copy-pasted at four sites each. #266 added the fourth copy of both,
which is the point at which this stopped being a coincidence.

**(a) The logger-resolution block is verbatim in four places.**

```csharp
try
{
    var logger = services?.GetService<ILoggerFactory>()?.CreateLogger(Category);
    logger?.LogWarning("…", …);
}
catch
{
    // Ignored: a failing diagnostic must not take the form down with it.
}
```

Present in `MaskedLinesDiagnostic.Warn`, `PasswordAdornmentDiagnostic.Warn`,
`MaskedValueDiagnostic.Warn`, and inline in `MudBlazorFieldComponentBase` — the *identical* comment
and empty `catch` included. Everything but the message template and its arguments is duplicated.

**(b) The once-per-field latch idiom is hand-written at four sites, one of them inverted.**

```csharp
ItemFieldScope?.ShouldWarnOnce(SomeDiagnostic.Category, DiagnosticFieldKey) ?? true
```

At `MudBlazorTextFieldComponent.razor.cs` (three sites: masked-lines, password-adornment,
masked-value) and at `MudBlazorFieldComponentBase.cs`, where the same rule is spelled **inverted** as
`… == false`. Two spellings of one invariant, and the null-coalescing default (`?? true` — "no
collection scope means always report") re-derived every time.

That inversion is the sharp edge. Get the polarity backwards and the diagnostic goes **silent outside
collections** — the common case — while still working inside them, so the tests most likely to exist
still pass. There is no test today that would catch it.

## Proposed solution

Express each idiom once.

- **`DiagnosticLog.Warn(services, category, template, args)`** — an internal static helper owning the
  resolve/create/log/swallow block. Each diagnostic type then holds only its `Category`, its `Applies`
  rule, and its message.
- **`protected bool ShouldReport(string category)`** on `MudBlazorFieldComponentBase` — one place that
  knows the latch is category-qualified and that "no scope" means report. Call sites become
  `if (Rule.Applies(…) && ShouldReport(Rule.Category))`.

Neither changes behaviour; both are covered by the existing diagnostic suites, which is what makes
this safe to do as a pure refactor.

## Alternatives considered

**Leave it.** Four copies is survivable, and each is individually correct today. Rejected because the
count only goes up — this repo adds diagnostics steadily (#181, #207, #212, #219, #266) and each new
one copies the idiom again. The inverted fourth copy of (b) is what a fifth author copies next.

**A base-class `Warn` method instead of a static helper.** Would fold (a) and (b) into one call. Tempting,
but the diagnostics are `static` types deliberately — their rules are pure and unit-testable without a
component. Making emission an instance method drags them back into the component's lifetime.

**An analyzer or a source generator.** Far too much machinery for four call sites.

## Area

`FormCraft.ForMudBlazor` — `Diagnostics/` and `Fields/`.

**Related:** #274, #266, #219, #207, #181. Distinct from #279, which moves UI-agnostic machinery across
the **package** boundary into core; this is duplication **within** `FormCraft.ForMudBlazor`. If #279
lands first, `DiagnosticLog` is a candidate to move with it — worth sequencing, not merging.

<details>

### Problem / context

The diagnostics family grew one issue at a time, and each new one was written by copying its nearest
neighbour — which is exactly why they are so consistent, and why the duplication went unnoticed. The
consistency is the tell: identical comments across four files are not independent convergence.

The latch idiom (b) is the higher-value half despite looking smaller. (a) is boilerplate — annoying,
low-risk, a change to it is a change in four places. (b) encodes an **invariant** (report unless a
collection scope has already reported this field for this category), and an invariant expressed four
times in two polarities is one bad copy away from a silent regression.

### Approaches

**A. Both helpers, one PR.** They touch the same call sites, so doing them together means reading each
site once. Cheap, and the existing suites cover the behaviour.

**B. Only the latch helper (b); leave the logging block.** Targets the actual risk and leaves the
harmless duplication. Smaller diff. But it revisits every call site anyway, so the marginal cost of
(a) is close to zero once you are there.

**C. Wait for #279.** It is already moving UI-agnostic machinery to core, and a logging helper is
plausibly UI-agnostic. Rejected as a *blocker*: the latch is MudBlazor-specific (it hangs off
`CollectionItemFieldScope`), so most of this work is not #279's, and waiting means the next diagnostic
copies the idiom a fifth time.

### Recommendation

**A**, with (b) first. Do the latch helper and add the test that pins its polarity — the test is the
part that has standing value independent of the refactor — then fold in the logging helper while the
call sites are open.

</details>

<details>

### Goal

Adding a fifth diagnostic requires writing its category, its rule and its message — and nothing else.

### Scope

- `DiagnosticLog.Warn(...)` — the shared emitter.
- `MudBlazorFieldComponentBase.ShouldReport(string category)` — the shared latch check.
- All four existing call sites converted to both.
- A test pinning the latch's polarity outside a collection.

### Non-goals

- Changing any diagnostic's rule, message text, category, or when it fires.
- Making the diagnostics public, or moving them to core (that is #279's boundary question).
- Adding event ids or an `IsEnabled` check — worth considering later, but a behaviour change, so not here.

### Behaviour

No observable change. Every existing diagnostic test must pass **unmodified** — that is the acceptance
criterion for the whole issue, and any test that needs editing signals a behaviour change to
investigate rather than accommodate.

### Key files

- create `FormCraft.ForMudBlazor/Diagnostics/DiagnosticLog.cs`.
- modify `Diagnostics/MaskedLinesDiagnostic.cs`, `Diagnostics/PasswordAdornmentDiagnostic.cs`, `Diagnostics/MaskedValueDiagnostic.cs`.
- modify `Fields/MudBlazorFieldComponentBase.cs` (add `ShouldReport`, convert its own inline copies).
- modify `Fields/TextField/MudBlazorTextFieldComponent.razor.cs` (three call sites).
- tests: `FormCraft.ForMudBlazor.UnitTests/Diagnostics/`.

### Validation rules

- `DiagnosticLog.Warn` never throws — including when `services` is null, when no `ILoggerFactory` is registered, and when the factory itself throws.
- `ShouldReport(category)` returns `true` when there is no `ItemFieldScope`, and otherwise defers to `ShouldWarnOnce(category, DiagnosticFieldKey)`.
- `ShouldReport` must be called **only after** the rule says yes — `ShouldWarnOnce` mutates the latch (#274).

### Edge cases

- **Polarity outside a collection.** The regression this issue is really about: a field with no `ItemFieldScope` must always report. Pin it with a test before refactoring, so the refactor is verified rather than trusted.
- **The category must stay part of the latch key.** One field can trip several diagnostics; a key without the category silences all but the first (`CollectionItemFieldScope.ShouldWarnOnce`'s own doc says so, and #274 added a test).
- **`ShrinkLabelDiagnosticCollector` is different** — it reports to a collector that dedupes by key, and only falls back to the latch when no collector exists. Convert it only if the fallback path is genuinely the same shape.

### Assumptions

- `DiagnosticLog` stays `internal`; the test project already has `InternalsVisibleTo` (#274).
- Message templates and their argument order are unchanged, so the existing `ShouldContain` assertions keep passing.

</details>

## 🛠️ Implementation plan

**Goal:** express the two diagnostic idioms once each, with no behaviour change.
**Architecture:** a static emitter beside the diagnostics; a protected latch check on the field component base.

**Global Constraints**

- **No behaviour change.** Every existing diagnostic test passes **unmodified**; a test that needs editing is a signal to stop and investigate.
- **The latch key keeps its category qualifier** — dropping it lets one diagnostic silence another on the same field.
- **Call the latch only after the rule says yes** — it has a side effect (#274).
- **`dotnet test --filter` is inert here** (`MTP0001`) — run `dotnet test -c Release`. (See #277.)

### Task 1: Pin the latch polarity before touching it

**Files:** test `FormCraft.ForMudBlazor.UnitTests/Diagnostics/MaskedValueDiagnosticTests.cs` (or a new `DiagnosticLatchTests.cs`).

**Interfaces:** none — characterisation only.

- [x] **Step 1:** Write tests asserting each of the four diagnostics still reports for a field **outside** any collection — the case an inverted latch would silently kill.
- [x] **Step 2:** Run `dotnet test -c Release` → PASS (they pass today; that is the point — this is the safety net, written first).
- [x] **Step 3:** Commit: `test(mudblazor): pin that diagnostics report outside a collection`

### Task 2: Extract the shared latch check

**Files:** modify `FormCraft.ForMudBlazor/Fields/MudBlazorFieldComponentBase.cs`, `Fields/TextField/MudBlazorTextFieldComponent.razor.cs`.

**Interfaces:** `protected bool ShouldReport(string category)` — `ItemFieldScope?.ShouldWarnOnce(category, DiagnosticFieldKey) ?? true`.

- [x] **Step 1:** Add `ShouldReport(string category)` to the base, documenting why the default is `true` and why the category is part of the key.
- [x] **Step 2:** Convert all four call sites, including the inverted `== false` one in the base.
- [x] **Step 3:** Run `dotnet test -c Release` → PASS with **no test edited**.
- [x] **Step 4:** Commit: `refactor(mudblazor): express the diagnostic latch check once`

### Task 3: Extract the shared emitter

**Files:** create `FormCraft.ForMudBlazor/Diagnostics/DiagnosticLog.cs`; modify the three diagnostic types and `Fields/MudBlazorFieldComponentBase.cs`; test `FormCraft.ForMudBlazor.UnitTests/Diagnostics/DiagnosticLogTests.cs`.

**Interfaces:** `internal static void Warn(IServiceProvider? services, string category, string template, params object?[] args)`.

- [x] **Step 1:** Write failing tests for `DiagnosticLog.Warn`: logs under the given category; does not throw with a null provider; does not throw with no `ILoggerFactory` registered; does not throw when the factory throws.
- [x] **Step 2:** Run `dotnet test -c Release` → FAIL (type does not exist).
- [x] **Step 3:** Implement `DiagnosticLog`, then rewrite the three `Warn` methods and the base's inline copy to delegate to it, leaving each holding only its template and arguments.
- [x] **Step 4:** Run `dotnet test -c Release` → PASS with **no existing test edited**.
- [x] **Step 5:** Commit: `refactor(mudblazor): express diagnostic logging once`

phmatray:
Implements #284.

Closes #284.

### Plan
- [x] Task 1: Pin the latch polarity before touching it
- [x] Task 2: Extract the shared latch check
- [x] Task 3: Extract the shared emitter

### What landed

- `DiagnosticLog.Warn(services, category, template, args)` — the resolve/log/swallow block, expressed once. The service resolution stays **inside** the guard (a disposed circuit throws on resolve, not returns null); `DiagnosticLogTests` pins all four arms.
- `MudBlazorFieldComponentBase.ShouldReport(string category)` — the latch check, expressed once, including the site that used to spell it inverted as `== false`.
- `DiagnosticLatchTests` — pins that all four diagnostics report **outside** a collection, which is what the `?? true` default encodes.

No behaviour change: **no pre-existing test was edited**, and all 1349 pass. Verified by mutation rather than assertion — flipping the new default to `?? false` fails 9 tests, 4 of them the new ones. The ShrinkLabel case is caught by **nothing but** the new suite, because inside a `FormCraftComponent` that diagnostic returns at the collector and never reaches the latch.

`ShrinkLabelDiagnosticCollector` is deliberately **not** converted: its emitter needs the logger as a *value* (it marks a batch logged even with nothing to log to), so it is not the same shape.

### Follow-ups

- **The `?? true` default means an ordinary field re-emits on every re-mount.** A conditionally-visible or wizard-step field is destroyed and re-created, so `OnInitialized` runs again and, with no `ItemFieldScope`, the latch is skipped — README's "once per field" only holds inside a collection. Pre-existing and preserved deliberately here (this PR is a no-behaviour-change refactor); the fix is a per-instance flag like `_shrinkLabelDiagnosticEmitted`.
- **`ThrowingServiceProvider` now exists twice** — the new shared one in `TestSupport/`, and the older private nested copy in `ShrinkLabelDiagnosticCollectorTests`. Folding the copies in is #205's job; the nested one shadows at its use site, so behaviour is unchanged.
- **`ShouldReport` is a `protected` member on a public base.** No collision in-repo (build is clean), but this file documents the CS0108-under-`TreatWarningsAsErrors` hazard for exactly this reason. The name is the one the issue's spec prescribes; worth revisiting if an external subclass ever collides.
- **`.worktrees/` is not in `.gitignore`** (`.claude/worktrees/` is, at `.gitignore:16`). Unrelated to this change; noted because the kit's worktree guard flags it.
- **The repo profile's "Worktree home" note is stale** — it claims `.claude/worktrees/` is not ignored because two entries collapsed onto one line. They are on separate lines (13 and 16) and both are ignored.


phmatray:
Recorded from the merge triage — an observation, not tracked work, so it lives here rather than in an issue.

**`ShouldReport` is a `protected` member on a public base.** `MudBlazorFieldComponentBase<TModel, TValue>` is public API, and this file documents the CS0108-under-`TreatWarningsAsErrors` hazard twice already (it is why `DiagnosticServiceProvider` is named the way it is rather than simply making the injected property protected). Nothing in this repo collides and the build is clean at 0 warnings, but `ShouldReport` is a generic enough name that an external subclass could already have one, and it was not scoped the way `DiagnosticServiceProvider` was.

Kept as-is deliberately: `ShouldReport(string category)` is the interface #284's spec prescribed, and renaming it on a naming preference alone was not the right call to make unilaterally at merge time. Worth revisiting only if a real collision turns up.

Issue #262, opened 2026-08-12
Mark required file-upload fields — the one field type #199 left unannounced


phmatray:
## Problem / motivation

#199 made a `.Required(...)` field announce itself to assistive technology on **every** field type that
can carry the annotation — text, numeric, date, select, multi-select, autocomplete, lookup, LOV and
boolean, on both render paths. File upload is the one exclusion, and it is pinned rather than
overlooked:

```csharp
// AriaRequiredTests
Required_FileUpload_Field_Should_Stay_Unannotated()
```

**Why it was excluded.** `MudFileUpload` accepts `Required` and would emit `aria-required` on its
`<input type="file">`. But FormCraft renders that input deliberately out of reach:

```razor
InputClass="absolute mud-width-full mud-height-full overflow-hidden z-10 opacity-0"
tabindex="-1"
```

It sits at `opacity-0` beneath a custom drop zone, removed from the tab order. The affordance a user
actually reaches is a `MudButton`, which takes no `aria-required`. Binding the flag would satisfy a DOM
assertion while telling no keyboard or screen-reader user anything — the "forwarded but inert" failure
that `RenderPipelineParityTests`' sibling assertions exist to catch.

**Why this is worth fixing now rather than never.** Before #199 no field carried a required marker, so
the absence of one on a file-upload field meant nothing. Now every other required field renders `*` and
`aria-required="true"`, so **absence reads as "optional"**. A form with a required passport-scan upload
beside asterisked text fields actively signals the upload as optional, to sighted and screen-reader
users alike. That is a stronger wrong signal than the uniform silence it replaced — the same reasoning
that made #199 extend coverage to every other type instead of stopping at text.

WCAG 2.1 **3.3.2 Labels or Instructions** (Level A) expects required fields to be identified. This one
is not.

## Proposed solution

Identify the requirement at the **label level**, where the user actually is, rather than on the hidden
input:

- Render the required marker in the field's own label (the `<MudText>` FormCraft already emits above the
  drop zone), so it is visible.
- Give the reachable affordance — the `MudButton` / drop zone — an accessible name or
  `aria-describedby` that carries the requirement, so it is announced.

Both file-upload components need it (`MudBlazorFileUploadFieldComponent` single, and
`MudBlazorMultipleFileUploadComponent`), and the existing pinning test is inverted to assert the new
behaviour.

## Alternatives considered

- **Bind `Required` on `MudFileUpload` anyway.** One line, and it would emit `aria-required` plus
[...]

Issue #298, opened 2026-08-13
Field components cache their configuration at init, so a reused instance renders the previous field's settings


phmatray:
## Problem / motivation

Every MudBlazor field component reads its configuration **once, in `OnInitialized`**, and never looks
at it again. Blazor reuses a component instance whenever the render-tree shape matches, so an instance
can be handed a *different* `Context` while those cached attributes still describe the field it was
first rendered for. From then on it renders — and now reports — against the wrong field's settings.

All 16 components under `FormCraft.ForMudBlazor/Fields/` do this. `MudBlazorTextFieldComponent` is the
clearest case:

```csharp
protected override void OnInitialized()
{
    // ...
    Mask = GetAttribute<string?>(TextMaskMap.AttributeName);
    MaskCleanDelimiters = GetAttribute(TextMaskMap.CleanDelimitersAttribute, false);
    MaskFactory = GetAttribute<Func<IMask>?>(TextMaskMap.MaskFactoryAttribute);
}
```

`OnParametersSet` refreshes the **value** (`FieldComponentBase` reloads it via `ShouldReloadValue()`)
but nothing refreshes the **configuration**. So the `.razor` binds `Mask="@GetMask()"`, which resolves
from a `Mask` string belonging to the previous field.

**Two ways to reach it, both ordinary:**

1. **Swapping `FormCraftComponent.Configuration`** on a live form — a wizard step, a mode toggle, any
   UI that renders a different form over the same component tree.
2. **Collection rows shifting.** `CollectionFieldComponent.razor` renders items with
   `@for (var i = 0; i < Items.Count; i++)` and **there is no `@key` anywhere in this repository**, so
   Blazor matches item components *positionally*. Remove or reorder a row and every component from
   that index down is re-parameterised with a different item's context.

**Why it is worth filing now.** This is pre-existing and, until recently, only affected rendering —
bad enough, but silent. #283 made it additionally reachable by `MaskedValueDiagnostic`, which now runs
from the external-model-change branch of `OnParametersSet`; a stale `Mask` means the warning can quote
a pattern the field no longer uses, or fire about a mask that was removed. A diagnostic that names the
wrong pattern is worse than no diagnostic, because it sends the developer to audit data against a rule
their form does not apply.

## Proposed solution

Give the base class one hook for "my field configuration changed", call it on init **and** whenever
the field identity changes, and move each component's attribute loading into it. Add `@key` to the
collection loop so row identity stops depending on position.

[...]

Issue #297, opened 2026-08-13
Nothing detects a suite re-declaring a collection-item shape the fixture provides


phmatray:
## Problem / motivation

Three issues have now attacked the same defect, and each was found by a human reading files rather than
by anything that fails a build.

- **#205** extracted `CollectionItemFixture` because every collection-item suite carried its own copy of
  the same model pairs, and the copies had drifted.
- **#258** (merged as #272) migrated eleven suites onto it and closed with a verification grep:
  ```
  grep -rn "private class \(OrderModel\|OrderItem\|BasketModel\|BasketLine\|AppointmentModel\|AppointmentSlot\)"
  ```
  That matched on **six type names, not on shape**. Any suite expressing the same shape under a different
  vocabulary was invisible to it.
- **#282** (PR #296) found the five suites #258's check could not see — `CredentialsModel`/`Credential`,
  `VaultModel`/`VaultEntry` (byte-identical copies under two names), and two already-drifted
  `MixedModel`/`MixedRow` pairs — and migrated them **by hand**, via a manual shape-based sweep.

That sweep is a one-off. Nothing stops a sixth suite from declaring a private collection-item model
tomorrow, and **the failure mode is silent**: a nested private copy shadows the namespace-scope fixture
model, so a half-finished migration still compiles and still passes. The compiler is not the check, and
right now neither is anything else.

**The gap is demonstrably still open after #282.** Two suites now express the *same* shape:

```csharp
// ShrinkLabelKeyCollisionTests.cs
private class TwoCollectionModel
{
    public List<NamedOrderItem> Items { get; set; } = new();
    public List<NamedOrderItem> Extras { get; set; } = new();
}

// CollectionInputTypeTests.cs
private class TwoCollectionModel
{
    public List<OrderItem> Contacts { get; set; } = new();
    public List<OrderItem> Suppliers { get; set; } = new();
}
```

Both model *two same-typed collections on one root*, and both exist to exercise form-wide diagnostics
keyed by collection name — the same question. The first one's XML doc still asserts *"two collections on
one model is this suite's own shape, and no other suite needs it"*, which is now false. Neither #258's
grep nor #282's manual sweep flagged this, because both looked for item types with a single scalar
member.

Three passes, three manual discoveries. The next one should be a failing test.

## Proposed solution

Add a **guard test** in `FormCraft.ForMudBlazor.UnitTests` that reflects over the test assembly and fails
[...]

PR #307, opened 2026-08-13
ci(build): enforce dotnet format so .editorconfig stops being aspirational (#301)

### LINKED ISSUE #301: The .editorconfig's style rules are enforced nowhere, so 574 violations have accumulated
## Problem / motivation

`.editorconfig` declares code-style rules at `warning` severity, and `Directory.Build.props` sets
`<TreatWarningsAsErrors>true</TreatWarningsAsErrors>`. Read together those two lines say the style
rules are enforced and that violating one fails the build.

Neither is true. Measured on `dev` at the #289 merge:

```
dotnet format FormCraft.sln --verify-no-changes
→ 574 diagnostics across 201 distinct files
```

| Diagnostic | Count | What it is |
|---|---:|---|
| `IDE0011` | 273 | add braces to single-statement `if` |
| `FINALNEWLINE` | 167 | missing final newline |
| `WHITESPACE` | 101 | indentation / spacing |
| `IDE0032` | 14 | use auto property |
| `IMPORTS` | 8 | `using` ordering |
| `IDE0019` | 6 | use pattern matching |
| `IDE0044` | 5 | make field readonly |

**The root cause is not "nobody ran the formatter".** `IDE*` analyzers do not run during
`dotnet build` unless `EnforceCodeStyleInBuild` is set, and that property is **absent** from
`Directory.Build.props`. `TreatWarningsAsErrors` therefore never sees a style warning — it only ever
governs compiler and package analyzer warnings. CI does not close the gap either: `ci.yml` runs
`./build.cmd Test` and `continuous.yml` runs `./build.cmd Pack`, and neither invokes `dotnet format`.

So the severities in `.editorconfig` are **aspirational — nothing anywhere reads them.** That is the
defect, and it is worth stating precisely because the file *looks* authoritative: a contributor
reading `csharp_prefer_braces = true:warning` (`.editorconfig:66`) alongside `TreatWarningsAsErrors`
reasonably concludes braces are mandatory, and 273 counter-examples in the tree say otherwise.

## Proposed solution

Close the gap in the direction of **enforcement in CI, not in the build**:

1. Add a `Format` target to the Nuke build wrapping `dotnet format --verify-no-changes`.
2. Land a one-off formatting pass, split into a whitespace-only commit and a code-style commit so the
   201-file diff is reviewable.
3. Wire `Format` into `ci.yml` as its own step, so the rules cannot drift again.
4. Correct `CLAUDE.md` and `.claude/skills/repo-profile.md`, which currently record that CI runs no
   format check.

Deliberately **not** `EnforceCodeStyleInBuild=true`: combined with `TreatWarningsAsErrors` that makes
a single missing brace break `dotnet build` mid-edit, which is hostile during ordinary development
and slows every local build with the extra analyzer pass. A CI gate prevents regression — which is
the actual goal — without that cost.

## Alternatives considered

- **`EnforceCodeStyleInBuild=true` + fix all 574.** The strictest option and the most consistent with
  this repo's `TreatWarningsAsErrors` stance. Rejected as the default for the reason above: it is a
  hard build break until the very last violation is fixed, and it makes every incremental build pay
  for the style pass. Reasonable to revisit once the tree is clean and the CI gate has held for a
  while — at that point it costs one property and no diff.
- **Relax `.editorconfig` to `silent`/`suggestion`.** Zero diff, and it makes the file honest by
  describing what the codebase actually does. Rejected because it resolves the contradiction by
  discarding the rules rather than applying them, and 94% of the violations are mechanically
  auto-fixable — the cost of keeping the rules is one formatting pass, not ongoing effort.
- **Do nothing.** Defensible: none of the 574 is a bug. Rejected because the gap actively misleads —
  and because every new file lands in whichever style its author guessed, so the drift compounds.

## Area

Build, CI and repo hygiene — `.editorconfig`, `Directory.Build.props`, `build/Build.cs`,
`.github/workflows/ci.yml`, and a formatting pass across `FormCraft`, `FormCraft.ForMudBlazor`,
`FormCraft.ForFluentUI`, `FormCraft.DemoBlazorApp` and the three test projects.

**Related:** #276, #279

---

<details>

### Problem & context

Found while merging #289 (the #279 adapter-machinery move). That PR's Step 9 ran the profile's
format/lint verify command and found violations in six of the files it had touched — all of them
pre-existing, none introduced by the PR. Widening the check showed the scale: 574 diagnostics, 201
files, in a repo whose `.editorconfig` declares those very rules at `warning` and whose build treats
warnings as errors.

The contradiction resolves once you know that `IDE*` analyzers are opt-in at build time
(`EnforceCodeStyleInBuild`), and that property is not set. Nothing is broken; a promise is simply
unkept. That makes this the same *shape* of defect as #276 — "the `.html` test report is promised but
unenforced" — in a different subsystem, which is why the two are linked rather than merged.

### Approaches

**A. Enforce in the build.** Set `EnforceCodeStyleInBuild=true`, fix all 574. Strongest guarantee and
most consistent with `TreatWarningsAsErrors`. Costs: a hard build break until the last fix lands, an
analyzer pass on every incremental build, and a very unpleasant edit loop (you cannot compile to run
one test while a brace is missing).

**B. Enforce in CI.** One-off formatting pass, then `dotnet format --verify-no-changes` as a CI step
via a Nuke `Format` target. Local builds stay fast and correctness-focused; regression is still
impossible to merge. Costs: a contributor learns about a style slip at CI rather than at build, and
the repo carries one more target. **Recommended.**

**C. Retreat.** Relax the `.editorconfig` severities to `silent`/`suggestion` so the file documents
reality. Zero diff, zero enforcement, honest. A fine answer if the maintainer does not actually want
these rules — but it throws away rules that cost one `dotnet format` invocation to satisfy.

### Recommendation

**Approach B.** The goal is that the tree matches its declared style and stays that way; a CI gate
achieves exactly that, and build-time enforcement buys only a faster failure at a real daily cost.

Two sequencing notes that matter more than the choice:

- **Split the pass by `dotnet format` subcommand, not by folder.** `dotnet format whitespace` covers
  `WHITESPACE` + `FINALNEWLINE` (268 diagnostics) and is purely mechanical — reviewable at a glance.
  `dotnet format style` covers the `IDE*` rules (306) and *does* change code shape: `IDE0032` rewrites
  a backing field into an auto property, `IDE0019` restructures a cast into a pattern match,
  `IDE0044` adds `readonly`. Those 25 deserve reading; the 281 brace/import fixes in the same batch do
  not. One commit each keeps the reviewable part small instead of hiding it in a 201-file diff.
- **Land it when few PRs are open.** A 201-file reformat conflicts with essentially every in-flight
  branch. The conflicts are trivial to resolve (take either side, re-run the formatter) but there are
  a lot of them.

⚠️ One constraint from #289 has **expired** and should not be carried forward: that PR declined to
reformat `FormCraft/Forms/Validation/DynamicFormValidator.cs` because doing so would have destroyed
git's 95% rename detection for the move it was making. The move has landed; the file is now ordinary
history and should be formatted with everything else.

</details>

<details>

### Goal

`dotnet format FormCraft.sln --verify-no-changes` exits clean on `dev`, and CI fails any PR that makes
it dirty.

### Scope

**In:** a Nuke `Format` target; a two-commit formatting pass over the solution; a `ci.yml` step
invoking it; corrections to `CLAUDE.md` and `.claude/skills/repo-profile.md`.

**Non-goals:** changing any `.editorconfig` *rule* (only enforcement changes; if a rule turns out to
be wrong, that is its own issue); `EnforceCodeStyleInBuild`; touching `CHANGELOG.md`; reformatting
generated output under `obj/`.

### Where the gap is

```mermaid
flowchart TD
    EC[".editorconfig<br/>csharp_prefer_braces = true:warning"] --> IDE["IDE0011 analyzer"]
    IDE -.->|"needs EnforceCodeStyleInBuild<br/>❌ not set"| BUILD["dotnet build<br/>TreatWarningsAsErrors=true"]
    IDE -->|"runs here"| FMT["dotnet format"]
    FMT -.->|"❌ no CI step invokes it"| CI["ci.yml → build.cmd Test"]

    BUILD --> V["574 violations survive"]
    CI --> V

    style V fill:#ffd7d7,stroke:#b02b2b
    style EC fill:#cfe6ff,stroke:#2b6cb0
```

Both paths from the rule to a gate are cut. This issue restores the right-hand one.

### The gate

`build/Build.cs` gains:

```csharp
Target Format => _ => _
    .DependsOn(Restore)
    .Executes(() => DotNet($"format {Solution.Path} --verify-no-changes --no-restore"));
```

Invoked from `ci.yml` as a step of its own — **not** as a `DependsOn` of `Test`, so a local
`./build.cmd Test` is unaffected.

### Behaviour rules

1. **Enforcement only.** No `.editorconfig` rule is added, removed, or re-severitied. If the pass
   reveals a rule the project does not want, that is a follow-up, not a silent edit here.
2. **The whitespace commit must be reviewable as mechanical** — `dotnet format whitespace` only. Any
   code-shape change appearing in it means the wrong subcommand ran.
3. **The 25 shape-changing fixes** (`IDE0032` 14, `IDE0019` 6, `IDE0044` 5) are read individually
   before the commit. `dotnet format` is not a refactoring tool and its fixers are occasionally
   surprising on partial classes and razor code-behind.
4. **Tests must pass unchanged** after each commit. A formatting pass that alters behaviour has
   stopped being a formatting pass.
5. **The profile is part of the deliverable.** `.claude/skills/repo-profile.md` currently states
   *"CI runs no `dotnet format` check"* under *CI gates*; leaving that stale re-creates this issue's
   root cause one layer up.

### Assumptions

- The maintainer wants the `.editorconfig` rules kept rather than relaxed (Approach C stays available
  and would close this issue differently).
- Razor `.razor` files are out of `dotnet format`'s reach for these rules; only `.cs` (including
  `.razor.cs`) is affected. Verify on the first run rather than assuming.

</details>

## 🛠️ Implementation plan

**Goal:** `dotnet format FormCraft.sln --verify-no-changes` exits clean, and CI keeps it that way.

**Architecture:** A Nuke `Format` target is the gate; two scoped formatting commits clear the backlog;
a `ci.yml` step makes it binding; the docs stop claiming the check does not exist.

**Tech Stack:** .NET 8 / .NET 10, NUKE, `dotnet format`, GitHub Actions.

### Global Constraints

- `TreatWarningsAsErrors=true` stays. **Do not** add `EnforceCodeStyleInBuild` — that is Approach A,
  explicitly not chosen.
- **No `.editorconfig` rule changes.** Enforcement only.
- Never hand-edit `CHANGELOG.md`; `dotnet test --filter` is inert under MTP — use the full run.
- Behaviour-preserving: the full suite must pass unchanged after every commit.
- Land this when few PRs are open; a 201-file diff conflicts with every in-flight branch.
- Commit identity: `git -c user.email=phmatray@gmail.com -c user.name="Philippe Matray"`.
- Conventional Commits; scope `ci` for the gate, `style` for the passes.

---

### Task 1: A `Format` target that fails on the current tree

**Files:** modify `build/Build.cs`.

**Interfaces:** `Target Format` depending on `Restore`, running
`dotnet format <solution> --verify-no-changes --no-restore`.

- [x] **Step 1:** Add the `Format` target to `build/Build.cs`, depending on `Restore` and **not** wired into `Test` or `Pack`.
- [x] **Step 2:** Run `./build.sh Format` → FAIL, and record the counts it reports (expected ~574 diagnostics across ~201 files).
- [x] **Step 3:** Confirm the target fails for the right reason — the output names `.cs` files with `IDE0011`/`WHITESPACE`/`FINALNEWLINE`, not a missing-tool or restore error.
- [x] **Step 4:** Run `./build.sh Test` → PASS, proving the new target did not disturb the existing pipeline.
- [x] **Step 5:** Commit: `ci: add a Format target that verifies dotnet format is clean`.

---

### Task 2: The mechanical whitespace pass

**Files:** modify every `.cs` file `dotnet format whitespace` touches (expected ~268 diagnostics).

**Interfaces:** none — no code shape changes.

- [x] **Step 1:** Run `dotnet format FormCraft.sln whitespace` and capture `git diff --stat`.
- [x] **Step 2:** Read the diff and confirm it is **only** indentation, spacing and final newlines. Any brace, property or `using` change means the wrong subcommand ran — revert and re-run.
- [x] **Step 3:** Run `dotnet build -c Release` → PASS with zero warnings (`TreatWarningsAsErrors` makes this strict).
- [x] **Step 4:** Run `dotnet test -c Release` → PASS, unchanged counts.
- [x] **Step 5:** Commit: `style: apply dotnet format whitespace across the solution`.

---

### Task 3: The code-style pass, with the shape-changing fixes read individually

**Files:** modify every `.cs` file `dotnet format style` touches (expected ~306 diagnostics).

**Interfaces:** none intended — `IDE0032`/`IDE0019`/`IDE0044` alter member shape but not behaviour.

- [x] **Step 1:** Run `dotnet format FormCraft.sln style --diagnostics IDE0011 IMPORTS` first — the 281 purely mechanical fixes — and skim the diff.
- [x] **Step 2:** Run `dotnet build -c Release && dotnet test -c Release` → PASS.
- [x] **Step 3:** Run `dotnet format FormCraft.sln style` for the remaining `IDE0032`/`IDE0019`/`IDE0044` (~25) and **read every hunk**: auto-property conversions and pattern-match rewrites are occasionally surprising on partial classes and `.razor.cs` code-behind.
- [x] **Step 4:** Revert by hand any fix that hurts readability or fights a documented intent, and note which in the PR description — a rule the codebase deliberately breaks is a follow-up, not something to force.
- [x] **Step 5:** Run `./build.sh Format` → PASS (or, if Step 4 reverted anything, narrow `.editorconfig` for that rule in a separate follow-up rather than leaving the gate red).
- [x] **Step 6:** Run `dotnet test -c Release` → PASS, unchanged counts.
- [x] **Step 7:** Commit: `style: apply dotnet format code-style fixes across the solution`.

---

### Task 4: Make the gate binding, and correct the docs

**Files:** modify `.github/workflows/ci.yml`, `CLAUDE.md`, `.claude/skills/repo-profile.md`.

**Interfaces:** a `./build.cmd Format` step in the `build-and-test` job.

- [x] **Step 1:** Add a `./build.cmd Format` step to `ci.yml`'s `build-and-test` job, before `./build.cmd Test` so a style failure reports fast.
- [x] **Step 2:** Update `.claude/skills/repo-profile.md` — its *CI gates* section states *"CI runs no `dotnet format` check"* and its *Format/lint verify* entry says the build is the only gate. Both are now false.
- [x] **Step 3:** Add a short note to `CLAUDE.md` recording that `.editorconfig` is enforced by the `Format` CI gate rather than by the build, and that `EnforceCodeStyleInBuild` was deliberately not used.
- [x] **Step 4:** Run `./build.sh Format && ./build.sh Test` → PASS.
- [x] **Step 5:** Add a `## 🎉 Unreleased` bullet to `README.md` (union hot-spot — keep it additive).
- [x] **Step 6:** Commit: `ci: enforce dotnet format in CI and correct the docs that denied it`.

phmatray:
Implements #301.

Closes #301.

code-review pass.

### Plan
- [x] Task 1: A `Format` target that fails on the current tree
- [x] Task 2: The mechanical whitespace pass
- [x] Task 3: The code-style pass, with the shape-changing fixes read individually
- [x] Task 4: Make the gate binding, and correct the docs

### ⚠️ Merge timing

#301's Global Constraints say to land this **when few PRs are open** — the formatting pass touches
~200 files and will conflict with every in-flight branch. That constraint governs *merging*, not
building, so this PR gets built now and the timing decision stays with whoever merges it. At the time
of writing there were 6 active branches; consider landing those first, then re-syncing this one.

The conflicts are trivial to resolve (take either side, re-run `dotnet format`) but there will be a
lot of them.

---

### What the implementation turned up

Three things the plan did not anticipate, all now documented in `CLAUDE.md` and the repo profile so
the next person does not rediscover them:

**1. ⚠️ `dotnet format` writes merge-conflict markers into multi-targeted source files.** All three
shipping projects are `net8.0;net10.0`, and a fix applied once per TFM can land as a literal
`<<<<<<< TODO: Unmerged change from project 'FormCraft(net10.0)'` block **inside the `.cs` file** —
code that does not compile. Hit twice here (`FieldRendererBase.cs`, then two Fluent files). Verify
mode never writes, so **the CI gate itself is unaffected**; it only bites when *applying*. Mitigations
now recorded: check `grep -rl '<<<<<<< TODO' --include='*.cs' .` after any apply run, and scope with
`--include <file>` when a whole-solution run misbehaves.

**2. One `dotnet format` run does not always reach a fixpoint.** After formatting a `field`-keyword
property, verify still reported three `WHITESPACE` diagnostics on it; a second run cleared them. So
"I formatted and the gate still fails" is not necessarily a gate bug.

**3. `IDE0032` now rewrites private backing fields into C# 14's `field` keyword — and silently drops
the XML docs that lived on the field.** On `MudBlazorFileUploadComponentBase.RequiredDescriptionId`
that deleted the `<remarks>` explaining a real correctness invariant (#262 / `CLAUDE.md`: the hint id
must be unique *per rendered instance*, because item fields render one hint per row). The rewrite is
value-identical, so no test could have caught the loss. The explanation has been restored onto the
property, and this is called out in `CLAUDE.md` as a hunk type to read rather than skim.

### Sync with `dev`

Merged `origin/dev` (3 commits, including #291's 5,600-line Fluent UI parity work). Git merged it
**cleanly with no conflicts** — but the merged code was written before this gate existed and brought
21 fresh violations, so it needed its own formatting pass (`style: format the Fluent UI code merged
from dev…`). That is the expected steady state for this branch until it lands: *every* re-sync will
need one.

Also verified while there: #291 correctly flipped `ValidateCollections="true"` on the Fluent
container when it implemented collection rendering, which is the trap flagged on #278.

## Follow-ups

- **Make the `Format` target refuse a tree containing `<<<<<<< TODO` markers.** Finding 1 above is a
  silent corruption that currently only a human notices; a two-line guard in the target would turn it
  into a failed build. Out of scope here because the gate's job this PR is to verify formatting, not
  to police the formatter.
- **Revisit `EnforceCodeStyleInBuild=true` once this gate has held for a few releases.** #301 rejected
  it as the *entry* point (a hard build break until the last violation is fixed, plus an analyzer pass
  on every incremental build). With the tree already clean, that objection is much weaker — it would
  then cost one property and no diff.

### ⚠️ The automated code-review pass did not complete — read this before approving

`implement-issue`'s Step 7 calls for a `code-review` pass over the branch. **It was attempted three
times and never returned findings**, so this PR has *not* had one:

1. First run destroyed uncommitted work — it does `git checkout <ref> -- .`, which reverted a
   formatting pass mid-flight (the hazard the skill itself documents). Stopped it, re-did the work,
   committed before retrying.
2. Second run wedged: no output for ~9 minutes and no reply to a direct status request.
3. Third run, scoped to just the five hand-written files, also failed to return.

Most likely environmental — a dozen agent sessions were running against this repo concurrently.

**What stands in for it:** a manual review of every non-mechanical hunk, which is where the three
findings above came from (the corruption, the deleted `<remarks>`, the non-convergence). The
formatting churn is tool output that `./build.cmd Format` verifies in CI. But a second opinion on the
hand-written delta — `build/Build.cs`, `ci.yml`, `FieldRendererBase.cs`, `FieldRendererService.cs`,
`MudBlazorFileUploadComponentBase.cs`, `DemoComponentBase.cs` — is genuinely missing, so please give
those a human read rather than assuming a bot already did.

### Verification at the ready-flip

- `dotnet build -c Release` — clean, zero warnings (`TreatWarningsAsErrors`).
- `dotnet test -c Release` — **1,561 passing**, 1 skipped, 0 failing (158 Fluent / 864 core / 539 MudBlazor).
- `dotnet format --verify-no-changes` — clean.
- CI on an earlier head: **`Run ./build.cmd Format: success`** followed by `Run ./build.cmd Test: success` —
  the new gate proven to work in CI, not just locally.
- Synced with `dev` twice during the run (through #323/#311); each sync needed its own formatting
  pass for the newly merged code, which is expected until this lands.


phmatray:
### Merge held — `dev` is red, and not because of this PR

A `/merge-pr` run stopped short of merging. **`dev`'s own CI is failing** on its tip (`65944e5`):

```
Test  Failed  // Projects that did not complete: FormCraft.ForMudBlazor.UnitTests
```

Reproduced locally on a merge of `origin/dev` into this branch — one failing test, and it belongs to
neither side of that merge on its own:

```
FormCraft.ForMudBlazor.UnitTests.Fields.CollectionItemShapeGuardTests
  .No_Suite_Should_Re_Declare_A_Collection_Item_Shape_The_Fixture_Provides

  offenders should be empty but had 2 items:
    FieldConfigurationRefreshTests.EqualityItemModel holds List<EqualityItem>, whose shape (String)
      is already a CollectionItemFixture item type
    FieldConfigurationRefreshTests.EqualityItemModel is a collection root whose composition (String)
      is already modelled by CollectionItemFixture.NamedOrderModel or OrderModel
```

**Cause: two PRs that never saw each other.**

- **#306** (`0ad0d49`, from #297) added `CollectionItemShapeGuardTests` — a guard that fails the build
  when a suite re-declares a collection-item shape the shared fixture already provides.
- **#308** (`947a3a5`, from #298) added `FieldConfigurationRefreshTests` with its own
  `EqualityItemModel` / `EqualityItem` — exactly such a re-declaration.

Each was green against the `dev` it branched from; together they are red. Nothing in #307 touches
either file.

**Why this PR was not merged anyway.** Merging into a red base would bury this PR's own signal and
make a repo-wide formatting change look like the thing that broke the build. The `Format` gate this
PR adds runs *before* `Test` in `ci.yml`, so a reader seeing red on `dev` afterwards would reasonably
suspect the new gate first.

**Where this branch stands:** at `1df9f0b`, synced through #328, working tree clean. Its last two
heads (`86c3222`, `fe2f586`) both passed CI — `Run ./build.cmd Format: success` followed by
`Run ./build.cmd Test: success`. It currently reads `DIRTY` only because `dev` has since taken #306
and #331.

**To land it:** fix `dev` first (the guard's own message names the remedy — add `EqualityItemModel`
to `FindOffenders`' allowlist with a reason, or switch that suite to the shared fixture), then re-run
`/merge-pr #307`. The re-sync will need one more formatting pass for whatever landed meanwhile, which
is expected and is the cost #301 flagged for landing a repo-wide format change in a busy repo.

PR #309, opened 2026-08-13
fix(mudblazor): dedupe field diagnostics across component re-mounts (#304)

### LINKED ISSUE #304: Diagnostics re-report when an ordinary field's component is re-mounted
## Problem / motivation

Follow-up from #294 (which implemented #284). Every diagnostic latch in `FormCraft.ForMudBlazor` lives
in state that a component **re-mount** destroys, so a field outside a collection re-reports the same
configuration warning every time its component is re-created.

Two kinds of latch exist today and neither survives:

- **Per-instance flags** — `_shrinkLabelDiagnosticEmitted` (`Fields/MudBlazorFieldComponentBase.cs`)
  and `_maskedValueReportingSettled` (`Fields/TextField/MudBlazorTextFieldComponent.razor.cs`). A new
  instance starts with fresh flags, by definition.
- **The collection scope latch** — `CollectionItemFieldScope.ShouldWarnOnce`, reached through
  `MudBlazorFieldComponentBase.ShouldReport`. It is cascaded by the *collection*, so it outlives a row's
  component — but an ordinary field has no `ItemFieldScope` at all, and `ShouldReport` then returns
  `true` unconditionally.

So the guarantee is asymmetric: **inside** a collection a field reports once however many rows render
and however often they churn; **outside** one, nothing dedupes across instances.

**How you reach it, both ordinary:**

1. `.WithVisibilityProvider(model => model.ShowDetails)` on a misconfigured field — every toggle of the
   controlling input destroys and re-creates the component, so twenty toggles emit twenty identical
   warnings about one field's configuration.
2. A wizard step revisited, or any UI that navigates away from and back to a form.

The warnings are noisy rather than wrong — each one accurately describes a real misconfiguration — but
volume is exactly what gets a diagnostic muted, and a muted diagnostic takes the real signal with it.
That is the reasoning #207, #219 and #266 each used when they added a latch in the first place.

## Proposed solution

Give the form the same thing a collection already has: **one diagnostic latch that outlives a
component**, cascaded by `FormCraftComponent<TModel>` and keyed by `(category, DiagnosticFieldKey)` —
the key `ShouldReport` already builds.

`ShouldReport` then resolves in three steps rather than two:

1. an `ItemFieldScope` if there is one (unchanged — a collection still dedupes per field across rows);
2. otherwise the **form-level** scope, when the field is inside a `FormCraftComponent`;
3. otherwise `true` — a genuinely standalone field, rendered through
   `IFieldRendererService.RenderField` with no form at all, has nothing to dedupe against and must
   report.

```mermaid
flowchart TD
    A["ShouldReport(category)"] --> B{ItemFieldScope?}
    B -- "yes: inside a collection" --> C["ShouldWarnOnce(category, DiagnosticFieldKey)"]
    B -- no --> D{FormDiagnosticScope?}
    D -- "yes: inside a form (NEW)" --> E["ShouldWarnOnce(category, DiagnosticFieldKey)"]
    D -- "no: standalone field" --> F["true — always report"]
```

Step 3 is the polarity `DiagnosticLatchTests` pins, and it stays exactly as it is.

⛔ **Do not 'fix' this by changing `ShouldReport`'s `?? true` default.** That default is deliberate
(#284): inverting it silences the diagnostics for ordinary fields while leaving them working inside
collections — the failure hides in the case least likely to be tested. The gap here is that there is
no form-scoped place to look *before* falling back to that default, not that the default is wrong.

## Alternatives considered

**Leave it (WONTFIX).** A legitimate outcome, and a maintainer should feel free to close this on it.
These are development-time diagnostics; the duplicates are noise, not incorrect behaviour, and they
only appear for a field that is *already* misconfigured. A form-scoped latch is real machinery — a new
cascading value, a new type, and a hook in `FormCraftComponent` — for a problem whose worst symptom is
a repeated console line in an app that has a bug anyway. **The cost may simply not be worth it.**

**A static/`AsyncLocal` latch.** Cheapest to write and wrong: it would leak across circuits and across
forms in a server-rendered app, silencing a genuine warning on a second user's form because a first
user's form already reported it.

**Reuse `ShrinkLabelDiagnosticCollector`.** It is already cascaded form-wide and already dedupes by
key, so it looks like the answer. Rejected: it is an *aggregating reporter* — it batches conflicts and
flushes them as one combined warning — not a latch the other three diagnostics can consult. Bending it
to serve both would give one type two jobs.

## Area

`FormCraft.ForMudBlazor` — `Fields/` and `Features/FormContainer/`.

**Related:** #284, #294, #298, #207, #219, #266, #283.

Distinct from **#298**, which is the closest neighbour and does **not** subsume this. #298 is about a
*reused* instance rendering and reporting against the **previous field's** cached configuration — a
correctness bug, fixed by re-resolving configuration when field identity changes (plus `@key` on the
collection loop). This is about a *re-created* instance re-reporting the **same field's** configuration
— a duplication problem, fixed by putting the latch somewhere that outlives the instance. #298's hook
fires within an instance and adds no cross-instance state, so landing it leaves this untouched. They
touch the same file and are worth sequencing; #298 should land first.

<details>

### Problem / context

The diagnostics family (#181, #207, #219, #266, #283) grew one issue at a time, and each new one
inherited its predecessor's latching approach. #284 unified *how* the latch is asked
(`ShouldReport`), which is what makes this gap visible as a single fact rather than four coincidences:
there is now exactly one line that decides, and it has no form-scoped case.

The asymmetry is the tell. A collection got a durable latch because the pain was obvious — fifty rows,
fifty warnings, immediately. An ordinary field's re-mount produces the same duplication spread over
time, so it was never as loud, and nobody built it a home.

### Approaches

**A. Form-scoped latch cascaded by `FormCraftComponent`.** Mirrors `CollectionItemFieldScope` exactly
— same `ShouldWarnOnce(category, key)` shape, same cascading mechanism, same key. `ShouldReport` gains
one fallback step. Standalone fields keep reporting always. Small, symmetric, and the two scopes can
share an interface so `ShouldReport` reads as one lookup with two providers.

**B. Per-instance flags only, made durable via a keyed cache on the model or context.** Avoids a new
cascading value by hanging the latch off something already reachable. Rejected: the model is the
user's own type and is not FormCraft's to annotate, and the render context is rebuilt per render.

**C. Do nothing.** See *Alternatives considered* — genuinely defensible.

### Recommendation

**A**, if it is judged worth doing at all. It reuses a shape the codebase already has rather than
inventing one, and it leaves #284's pinned polarity untouched by adding a step *before* the default
rather than changing the default. Sequence it after **#298**, which is a correctness bug in the same
file and should not have to merge around this.

</details>

<details>

### Goal

A field's configuration diagnostic is emitted **once per field per form**, whether or not the field
lives in a collection, and regardless of how many times its component is destroyed and re-created.

### Scope

- A form-level diagnostic scope type, cascaded by `FormCraftComponent<TModel>`.
- `MudBlazorFieldComponentBase.ShouldReport` consults it when `ItemFieldScope` is null.
- Tests for the re-mount case (toggle a visibility-controlled field, assert one warning).

### Non-goals

- Changing any diagnostic's rule, message, category, or when it fires for the first time.
- Changing the standalone (no-form, no-collection) behaviour — it must keep reporting always, per
  `DiagnosticLatchTests`.
- Touching `ShrinkLabelDiagnosticCollector`'s aggregation.
- #298's configuration-staleness problem.

### Behaviour

| Placement | Before | After |
|---|---|---|
| Item field in a collection | once per field | once per field (unchanged) |
| Ordinary field in a form, re-mounted N times | N warnings | once per field |
| Standalone field, no form | always reports | always reports (unchanged) |

### Key files

- create the scope type beside `CollectionItemFieldScope` in `Fields/MudBlazorFieldComponentBase.cs`
  (or its own file), plus a `FormCraftCascadingValues` name for it.
- modify `Features/FormContainer/FormCraftComponent.razor` — cascade one instance per form.
- modify `Fields/MudBlazorFieldComponentBase.cs` — `ShouldReport`'s fallback chain.
- tests: `FormCraft.ForMudBlazor.UnitTests/Diagnostics/DiagnosticLatchTests.cs`.

### Validation rules

- `ShouldReport` returns `true` when there is neither an item scope nor a form scope.
- The category stays part of the key — one field tripping two diagnostics must report both (#274).
- The latch is still consulted **only after** the diagnostic's rule says yes; it mutates.
- One scope instance per form, not per render.

### Edge cases

- **Two forms over the same model.** Separate scope instances, so each reports — they are separate
  forms and a developer looking at one should not have its warning suppressed by the other.
- **A field both in a collection and re-mounted.** The item scope wins and already covers it.
- **Standalone `IFieldRendererService.RenderField`.** No form, no cascade, still reports — this is the
  case `DiagnosticLatchTests` pins and it must not change.

### Assumptions

- The form scope may be `internal`; the test project has `InternalsVisibleTo`.
- `FormCraftComponent` is the only place a form-wide cascade can originate, as it is for
  `ShrinkLabelDiagnosticCollector` today.

</details>

## 🛠️ Implementation plan

**Goal:** one diagnostic latch per field per form, surviving component re-mounts.
**Architecture:** a form-scoped latch cascaded by `FormCraftComponent`, consulted by `ShouldReport` between the collection scope and the report-always default.

**Global Constraints**

- **⛔ Do not change `ShouldReport`'s final `?? true`.** Add a step before it. The polarity is pinned by `DiagnosticLatchTests` and inverting it is the silent-outside-collections regression #284 exists to prevent.
- **The latch key keeps its category qualifier** — dropping it lets one diagnostic silence another on the same field.
- **Call the latch only after the rule says yes** — it has a side effect (#274).
- **Sequence after #298** — it rewrites configuration loading in the same components; landing this first would force it to merge around this.
- **`dotnet test --filter` is inert** (`MTP0001`) — run `dotnet test -c Release` (see #277 for the per-class form that does work).

### Task 1: Pin the re-mount duplication as a failing test

**Files:** test `FormCraft.ForMudBlazor.UnitTests/Diagnostics/DiagnosticLatchTests.cs`.

**Interfaces:** none — characterisation only.

- [x] **Step 1:** Add a test rendering a misconfigured field inside a `FormCraftComponent` behind `.WithVisibilityProvider(...)`, toggling it off and on three times, asserting exactly one warning.
- [x] **Step 2:** Run `dotnet test -c Release` → FAIL (three warnings today), confirming the gap is real before building machinery for it.
- [x] **Step 3:** Commit: `test(mudblazor): pin that a re-mounted field reports its diagnostic once`

### Task 2: Add the form-scoped latch and cascade it

**Files:** modify `FormCraft.ForMudBlazor/Fields/MudBlazorFieldComponentBase.cs` (new scope type + `FormCraftCascadingValues` entry), `Features/FormContainer/FormCraftComponent.razor`.

**Interfaces:** a form-scoped latch exposing `bool ShouldWarnOnce(string category, string key)`, cascaded under a new `FormCraftCascadingValues` name; one instance per form.

- [x] **Step 1:** Add the scope type, documenting that it is the *form's* latch and how it differs from `CollectionItemFieldScope` (which is per collection and survives row churn).
- [x] **Step 2:** Cascade one instance from `FormCraftComponent`, created once per component rather than per render.
- [x] **Step 3:** Run `dotnet test -c Release` → still FAIL on Task 1's test (nothing consults it yet), PASS everywhere else.
- [x] **Step 4:** Commit: `feat(mudblazor): cascade a form-scoped diagnostic latch`

### Task 3: Consult it from ShouldReport

**Files:** modify `FormCraft.ForMudBlazor/Fields/MudBlazorFieldComponentBase.cs`.

**Interfaces:** `ShouldReport(string category)` → item scope, else form scope, else `true`.

- [x] **Step 1:** Add the fallback step, keeping the final `?? true` for the no-scope-at-all case and documenting all three arms.
- [x] **Step 2:** Run `dotnet test -c Release` → PASS, including Task 1's test and the existing standalone-polarity tests **unmodified**.
- [x] **Step 3:** Verify by mutation: temporarily make the form-scope arm return `false` unconditionally and confirm the standalone tests still pass while the new one fails — proving the arms are independent. Revert.
- [x] **Step 4:** Commit: `fix(mudblazor): dedupe field diagnostics across component re-mounts`

phmatray:
Implements #304.

Closes #304.

### Plan
- [x] Task 1: Pin the re-mount duplication as a failing test
- [x] Task 2: Add the form-scoped latch and cascade it
- [x] Task 3: Consult it from ShouldReport

### What landed

- **`FormDiagnosticScope`** — cascaded by `FormCraftComponent` as a `readonly` field with `IsFixed="true"`, so it outlives the field components inside it.
- **`ShouldReport` now has three arms** — `ItemFieldScope`, then `FormDiagnosticScope`, then `return true`. The last arm is byte-for-byte the behaviour #284 pinned; only the middle one is new.
- **`DiagnosticOnceLatch`** — the `HashSet` + `$"{category}|{key}"` key format, extracted so the second scope reuses it rather than planting the fifth copy of the idiom #284 existed to remove. `CollectionItemFieldScope` delegates to it; its public signature and behaviour are unchanged.

Measured rather than asserted: before the fix the re-mount test saw **4** warnings (1 + 3 cycles), after it **1**. A mutation making the form arm always deny fails 13 tests while all four standalone-polarity tests stay green — the arms are independent, so #284's pinned `true` cannot be broken by this one.

### Scope limits — what this does *not* fix

- **A re-mounted *form* starts over.** The latch is a field on `FormCraftComponent`, so anything destroying the form takes it along. A wizard that puts a **separate form in each step** — which `StepperForm.razor` in the demo app does — is therefore not covered if the step host unmounts inactive content. The doc comments now say this rather than claiming "a wizard step revisited" outright, which was an unverified claim in the first draft.
- **Item fields inside a collection still re-report when the *collection* re-mounts.** `ShouldReport` takes the item arm, and `CollectionItemFieldScope` is rebuilt by `CollectionFieldComponent.OnParametersSet`; flipping `CollectionFieldConfiguration.IsVisible` re-mounts it. Behaviour is identical before and after this PR, and fixing it would collide with the deliberate "don't burn two latches" precedence rule — but #304 should not be read as closing that case.

### Follow-ups

- **Two non-discriminating assertions predate this branch** at `DiagnosticLatchTests.cs:72-73`: `ShouldContain("Password")` and `ShouldContain("masked")` both match `MaskedLinesDiagnostic`'s own message template, so neither can fail. The two tests added here originally inherited the pattern and were fixed; the #284-era ones were left alone to keep this diff to its plan.

### Sequencing note

#304's own plan says to sequence it after **#298** (field components caching configuration at init). #298 is still open with no PR. Proceeding anyway because the two changes are additive to different lines — #298 rewrites attribute *loading* in `OnInitialized`, this adds a cascaded scope and one fallback branch in `ShouldReport` — but #298 should still land first if both are in flight.


phmatray:
Recorded from the merge triage — an observation about a deliberate scope limit, not tracked work, so it lives here rather than in an issue.

**Item fields inside a collection still re-report when the *collection* re-mounts.** `ShouldReport` takes the `ItemFieldScope` arm for them, and `CollectionItemFieldScope` is rebuilt by `CollectionFieldComponent.OnParametersSet` whenever that component is re-created — so flipping `CollectionFieldConfiguration.IsVisible` re-mounts the collection and its item fields report again. Same for a re-mounted **form**: the latch is a field on `FormCraftComponent`, so a wizard placing a separate form in each step (as `StepperForm.razor` does) starts over per step if the host unmounts inactive content.

Neither is a regression — behaviour is identical before and after #309 — and both are documented in the PR body and in `FormDiagnosticScope`'s XML docs. Not filed as an issue deliberately: #304 was already borderline-WONTFIX by its own admission (development-time diagnostics, duplicates are noisy rather than wrong), and these remaining cases are strictly narrower than the one just fixed. Fixing the collection case would also collide with the deliberate "don't burn two latches" precedence rule in `ShouldReport`.

If the duplicate warnings ever show up in a real app rather than in reasoning about one, that is the signal to file it — with the reproduction attached.

PR #311, opened 2026-08-13
test(ci): make the step guards match only the wiring they assert on (#302)

### LINKED ISSUE #302: Ci guards can still be satisfied by text that isn't the wiring they check
## Problem / motivation

Two of the `Ci/` guards can be satisfied by text that is **not** the wiring they mean to assert on, so
they read correct while checking nothing. #267 (PR #287) tightened one of the two step primitives; these
are the residual gaps in the same shared reader, and the first one sits on the release path.

### (a) The login gate is satisfied by a code comment

`TrustedPublishingWorkflowTests.ReleaseWorkflow_Should_Gate_The_Login_On_A_Created_Release` asserts
`ShouldContain("release_created")` over the login step's **unstripped** slice. That slice
(`release-please.yml` lines 146-169) contains `release_created` **twice**:

| Line | Text | Kind |
|---|---|---|
| 148 | `if: ${{ needs.release-please.outputs.release_created == 'true' }}` | the wiring being asserted |
| 153 | ``# The run step is split in two on `release_created`, and the split is the guarantee (#221).`` | **prose** |

So deleting the step's `if:` outright leaves the test **green** off the comment alone. That deletion is
the exact #226 regression this test exists to catch, and #226's whole point is what it costs: the login
is skipped, `steps.login.outputs.NUGET_API_KEY` is empty, Nuke's `Publish` is gated off — while
release-please has already cut the tag, the changelog and the GitHub Release. A **green run on a version
that never reached nuget.org**, on a path this repo exercises a handful of times a year.

Verified by extracting the real slice with the current algorithm.

This is **pre-existing** — the prose sat inside the slice before #267 moved the slice's start, so #287
did not introduce it. #267 left it standing because narrowing the assertion changes *what* the suite
asserts, which its Non-goals ruled out; PR #287 instead updated the class remark to state the defect
honestly rather than repeat the old claim that the unstripped slice is "harmless".

### (b) `StepNamed` is the looser primitive, on the weaker guarantee

`WorkflowSource.StepNamed`/`TryStepNamed` still take `Array.FindIndex`'s **first** match with no
ambiguity check, while `StepWithId` gained both an exact anchored match and a loud ambiguity failure in
#267. That is backwards with respect to what YAML actually guarantees:

- step **ids** are unique per job — so a duplicate is a defect, and #267 made it loud;
- step **names** need not be unique at all — so a duplicate is *legal*, and it silently resolves to the
  first.

The primitive with the weaker guarantee got the looser treatment. Concrete risk: if a job ever splits
its upload in two — a natural consequence of #256's per-project results directories — then
`TestReportingTests`' four content assertions read only the first step, and the second can silently
lack `if: always()`, point `path:` at the stale `**/TestResults/` glob, or drop
`if-no-files-found: ignore`, all while green.

## Proposed solution

Close both on the same principle #267 established — an assertion may only match the thing it claims to
be about.

1. Have `TrustedPublishingWorkflowTests`' private `StepWithId` helper read
   `WorkflowSource.Stripped(workflowFile)` instead of `WorkflowSource.Read(workflowFile)`, so no
   assertion in that suite can be answered by prose. Pin it with a test asserting the slice the suite
   reads carries no comment lines, and that `release_created` appears in it exactly once.
2. Give `StepNamed`/`TryStepNamed` the same all-matches-then-count shape `StepWithIdOrNull` already
   uses, so two steps sharing a name fails loudly instead of resolving to the first.

## Alternatives considered

- **Assert (a) via `StepCondition` instead of stripping.** `ReleaseWorkflow_Should_Not_Gate_The_Login_On_The_NuGetUser_Secret`
  already reads the `if:` line alone this way, and it is immune for exactly that reason. Narrower and
  correct — but it fixes the one assertion rather than the class, leaving every *other* claim in the
  suite still answerable by a comment. Stripping fixes the class; doing both is fine.
- **Strip inside `WorkflowSource.StepNamed`/`StepWithId` themselves.** Tempting, and wrong: the scope is
  the caller's to choose, `WithoutComments` exists so a documentation-only edit cannot redden the suite,
  and `TestReportingTests` deliberately strips *before* splitting jobs so a commented-out `./build.cmd`
  cannot mark a job as running the build. Forcing the policy into the primitive removes a choice two
  callers make differently on purpose.
- **Adopt a YAML parser.** Explicitly rejected by #267, and this issue does not meet its stated
  criterion for reopening — that the assertions turn *materially structural* (`needs:` graphs, matrix
  expansion, anchor/alias resolution). These are text-precision fixes; the reader stays hand-rolled.
- **Leave (b).** No workflow in the repo has duplicate step names today, so it is latent. Rejected on
  the same reasoning #267 used for the id path: the whole premise is that an approximation nobody has
  tightened drifts, and the asymmetry is now visible in one file.

## Area

Test infrastructure — `FormCraft.UnitTests/Ci/`: `WorkflowSource.cs`,
`TrustedPublishingWorkflowTests.cs`, `WorkflowSourceTests.cs`. No library code, no workflow files, no
`build/Build.cs` change, no new package.

**Related:** #267 (landed as PR #287; tightened the sibling primitive), #226 (the regression (a) is
supposed to catch), #221 (wrote the comment that satisfies it), #255 (created the shared reader), #198
(owns the Trusted Publishing rehearsal and plans OIDC regression tests — one of its guards is currently
inert), #256 (per-project results directories, the trigger for (b)).

<details>

### Problem / context

`FormCraft.UnitTests/Ci/` asserts on repository *files* rather than library behaviour, because the
things it guards run a handful of times a year (a release, a publish) or fail silently (#231's inert
reporting). #255 consolidated those suites onto one reader; #267 made its `id:` scan precise. What
neither addressed is the other half of precision: not *which step* an assertion binds to, but *which
text within that step* is allowed to satisfy it.

`WithoutComments` exists precisely because these files carry comment blocks longer than the wiring they
describe — `release-please.yml`'s login step has six lines of `if:`/`uses:`/`with:` and roughly
seventeen lines of prose around it. A whole-slice `ShouldContain` over that is far more likely to hit
prose than a reader would guess, and (a) is the measured case.

### Approaches

**A. Strip comments at the two call sites, and mirror `StepWithId`'s counting into `StepNamed`.**
*Trade-offs:* small, local, no new concepts — reuses the exact shape #267 built. Keeps the deliberate
per-caller choice about stripping. Does not make the primitives structurally incapable of the mistake;
a future caller can still pass raw text to a whole-slice `ShouldContain`.

**B. Make the primitives always strip.**
*Trade-offs:* structurally prevents (a) everywhere. But it deletes a choice two callers make
differently on purpose, and it would break `TestReportingTests`' reason for stripping *before* the job
split. Rejected.

**C. Assert only on specific keys — never on whole slices.** Give the suite a `StepKey(step, "if")`
accessor and forbid whole-slice `ShouldContain`.
*Trade-offs:* the strongest end state, and where this file arguably belongs. But it is a rewrite of
every assertion in two suites, and `ShouldMatch(@"uses:\s*NuGet/login@[0-9a-f]{40}\b")` is legitimately
a whole-slice regex claim. Too big for the defect at hand; worth revisiting if the suite grows.

### Recommendation

**A.** It fixes the measured defect and the visible asymmetry with the same shape the codebase just
adopted, in one file plus its tests. C is the better architecture and a much larger change; nothing here
blocks it later.

**Assumptions:** both suites stay green throughout — this tightens *what may satisfy* an assertion, not
what is asserted, so a red bar is a finding rather than a formatting problem. `release-please.yml`
remains the fixture. Stripping does not disturb the digest assertion, because `WithoutComments` drops
only whole-line comments and `uses: NuGet/login@8d19675… # v1.2.0` is a trailing comment.

</details>

<details>

### Goal

No `Ci/` guard can be satisfied by prose, and neither step primitive can silently resolve a duplicate.

### Scope

- `FormCraft.UnitTests/Ci/TrustedPublishingWorkflowTests.cs` — the private `StepWithId` helper reads
  stripped text; the class remark drops the defect disclaimer #287 added.
- `FormCraft.UnitTests/Ci/WorkflowSource.cs` — `StepNamed`/`TryStepNamed` count all matches.
- `FormCraft.UnitTests/Ci/WorkflowSourceTests.cs` — pin the duplicate-name edge.

### Non-goals

- No new NuGet dependency; no YAML parser (#267's rejection stands, and its reopening criterion is not met).
- No change to any workflow, to `build/Build.cs`, or to library code.
- Not job-scoping `TrustedPublishingWorkflowTests` — that is #198's.
- Not converting whole-slice assertions to per-key accessors (approach C above).

### Surface

```
internal static class WorkflowSource
    string? TryStepNamed(string scope, string stepName)   // unchanged signature; duplicates now fail loudly
    string  StepNamed   (string scope, string stepName, string? scopeDescription = null)
```

No signature changes. `StepNamed`'s absence message stays byte-identical; the ambiguity message mirrors
`StepWithId`'s wording with `named '<name>'` in place of `` `id: <id>` ``.

### Matching rules

```mermaid
flowchart TD
    A[scope text] --> B["find lines whose trimmed text starts with <br/>- name: '&lt;stepName&gt;'"]
    B --> C{how many?}
    C -->|0| D[return null]
    C -->|2+| E[fail loudly: ambiguous name]
    C -->|1| F[slice from that line to the next sibling item]
```

The matched line is already the step's own list item, so — unlike the id path — no walk-back is needed;
only the count changes.

### Edge cases

- Two jobs each with a step of the same name, whole file passed as scope: **legal**, and it fails here.
  That is the same whole-file-scope limitation the id path documents and #198 owns; `TestReportingTests`
  is already job-scoped and unaffected.
- A step whose name is a prefix of another (`Run: Test` vs `Run: Test extras`): unaffected — the match
  includes the closing quote (`- name: 'Run: Test'`).
- Stripping removes whole-line comments only, so `uses: … # v1.2.0` survives and the digest regex holds.

### Assumptions

`release_created` appears exactly once in the login step once comments are stripped (the `if:` line),
which is what makes the count assertion in Task 1 a precise pin rather than a magic number.

</details>

## 🛠️ Implementation plan

**Goal:** no `Ci/` guard can be satisfied by prose, and neither step primitive silently resolves a duplicate.

repository files (`.github/workflows/*.yml`, `build/Build.cs`) rather than on library behaviour.

**Global Constraints**

- `TreatWarningsAsErrors=true` is deliberate — never relax it to make a build pass. `TryStepNamed`
  returns `string?`; the asserting wrapper must not leak that.
- **No new NuGet package.** A YAML parser is out of scope (#267 closed that question).
- No change to any workflow, to `build/Build.cs`, or to library code. If an assertion would require
  editing a workflow to pass, stop — that is a finding, not a refactor.

### Task 1: Stop the login-step assertions from being satisfied by prose

**Files:** modify `FormCraft.UnitTests/Ci/TrustedPublishingWorkflowTests.cs`.

**Interfaces:** no public change. The private helper
`StepWithId(string workflowFile, string stepId)` keeps its signature; only the text it reads changes.

- [x] **Step 1:** Add the failing test to `TrustedPublishingWorkflowTests.cs`, asserting the slice the
      suite actually reads carries no prose and mentions the gate token exactly once:

```csharp
[Fact]
public void ReleaseWorkflow_Login_Step_Should_Be_Read_Without_Its_Prose()
{
    // A whole-slice ShouldContain cannot tell wiring from a comment, and this step is surrounded by
    // more prose than wiring — so `release_created` appearing twice is not a curiosity, it is the
    // reason ReleaseWorkflow_Should_Gate_The_Login_On_A_Created_Release passes when the `if:` is gone.
    var step = StepWithId("release-please.yml", "login");

    step.Split('\n').ShouldAllBe(l => !l.TrimStart().StartsWith("#", StringComparison.Ordinal));
    step.Split('\n').Count(l => l.Contains("release_created", StringComparison.Ordinal)).ShouldBe(1);
}
```

- [x] **Step 2:** Run `dotnet test -c Release` → **RED** (the slice carries `release-please.yml`'s
      lines 153-169, so the comment assertion fails and the count is 2).
- [x] **Step 3:** Change the private helper to read comment-stripped text:

```csharp
private static string StepWithId(string workflowFile, string stepId) =>
    WorkflowSource.StepWithId(WorkflowSource.Stripped(workflowFile), stepId, workflowFile);
```

- [x] **Step 4:** Run `dotnet test -c Release` → **GREEN**. Confirm
      `ReleaseWorkflow_Should_Exchange_The_OidcToken_For_A_ShortLived_Key` still passes — its
      `ShouldMatch(@"uses:\s*NuGet/login@[0-9a-f]{40}\b")` and `ShouldContain("NUGET_USER")` both read
      lines that survive whole-line stripping.
- [x] **Step 5:** Prove the guard is no longer inert: temporarily delete the `if:` line from the login
      step in `release-please.yml`, confirm
      `ReleaseWorkflow_Should_Gate_The_Login_On_A_Created_Release` now **fails**, then revert the
      workflow and re-run to green.
- [x] **Step 6:** Rewrite the class remark that PR #287 added — it documents the defect as standing, so
      it must now describe the fix instead: the suite reads stripped text, so an assertion cannot be
      answered by prose. Keep the note that `StepCondition`-based tests read the `if:` line alone.
- [x] **Step 7:** Commit: `test(ci): read the login step without its prose so the release gate is real`

### Task 2: Make a duplicate step name fail loudly

**Files:** modify `FormCraft.UnitTests/Ci/WorkflowSource.cs`,
`FormCraft.UnitTests/Ci/WorkflowSourceTests.cs`.

**Interfaces:** unchanged signatures — `string? TryStepNamed(string scope, string stepName)` and
`string StepNamed(string scope, string stepName, string? scopeDescription = null)`. The absence message
stays byte-identical; a new ambiguity message mirrors `StepWithId`'s.

- [x] **Step 1:** Add the failing test to `WorkflowSourceTests.cs`:

```csharp
[Fact]
public void StepNamed_Should_Fail_Loudly_When_Two_Steps_Share_A_Name()
{
    // Step names need not be unique in GitHub Actions — unlike ids — so this is the primitive with the
    // WEAKER guarantee, and it was the one still resolving a duplicate to the first silently.
    const string Steps = """
                         steps:
                           - name: 'Publish: test-results'
                             if: always()
                             uses: actions/upload-artifact@v4
                           - name: 'Publish: test-results'
                             uses: actions/upload-artifact@v4
                         """;

    Should.Throw<ShouldAssertException>(() => WorkflowSource.StepNamed(Steps, "Publish: test-results", "the fixture"))
        .Message.ShouldContain("ambiguous");

    Should.Throw<ShouldAssertException>(() => WorkflowSource.TryStepNamed(Steps, "Publish: test-results"))
        .Message.ShouldContain("ambiguous");
}
```

- [x] **Step 2:** Run `dotnet test -c Release` → **RED** (both calls return the first step instead of
      throwing).
- [x] **Step 3:** Reshape `TryStepNamed` to count every match, mirroring `StepWithIdOrNull`. Add a
      private core so the asserting form can still name its scope in the ambiguity message:

```csharp
internal static string? TryStepNamed(string scope, string stepName) =>
    StepNamedOrNull(scope, stepName, null);

private static string? StepNamedOrNull(string scope, string stepName, string? scopeDescription)
{
    var lines = scope.Split('\n');
    var header = $"- name: '{stepName}'";
    var matches = new List<int>();

    for (var index = 0; index < lines.Length; index++)
    {
        if (lines[index].TrimStart().StartsWith(header, StringComparison.Ordinal))
        {
            matches.Add(index);
        }
    }

    matches.Count.ShouldBeLessThan(
        2,
        $"{scopeDescription ?? "the searched text"} has {matches.Count} steps named '{stepName}' — an ambiguous name cannot be resolved to one step");

    return matches.Count == 0 ? null : StepFrom(lines, matches[0]);
}
```

- [x] **Step 4:** Point `StepNamed` at the core so its scope description reaches the ambiguity message,
      leaving its absence message byte-identical:

```csharp
internal static string StepNamed(string scope, string stepName, string? scopeDescription = null)
{
    var step = StepNamedOrNull(scope, stepName, scopeDescription);

    return step.ShouldNotBeNull(
        $"{scopeDescription ?? "the searched text"} no longer has a step named '{stepName}'");
}
```

- [x] **Step 5:** Update the `TryStepNamed` remark to record that ambiguity stays loud here for the same
      reason it does on the id path — absence and ambiguity are different answers, and reporting a
      duplicated step as a missing one would undo the tightening one call further out.
- [x] **Step 6:** Run `dotnet test -c Release` → **GREEN** with `TestReportingTests` unchanged (each job
      has exactly one `Publish: test-results` step today, so nothing becomes ambiguous).
- [x] **Step 7:** Commit: `test(ci): fail loudly when two steps share a name`

phmatray:
Implements #302.

Closes #302.

Makes two `Ci/` guards match only the wiring they claim to assert on. Test infrastructure only —
`FormCraft.UnitTests/Ci/`. No library code, no workflow files, no `build/Build.cs` change, no new
package.

All plan tasks landed; every checkbox on the issue is ticked.

### Plan
- [x] Task 1: Stop the login-step assertions from being satisfied by prose
- [x] Task 2: Make a duplicate step name fail loudly

### ⚠️ The issue's stated severity was corrected mid-implementation

#302's body says the inert login-gate test risks *"a green run on a version that never reached
nuget.org"*. **The inert test is real; that consequence was overstated**, and
[a comment on the issue](https://github.com/phmatray/FormCraft/issues/302) records the measurement.

Measured against the real `release-please.yml` with the pre-fix helper: with the login step's `if:`
**deleted**, and separately **re-pointed** at an unrelated condition,
`ReleaseWorkflow_Should_Gate_The_Login_On_A_Created_Release` passed both times — but the **suite still
went red** via `ReleaseWorkflow_Should_Not_Gate_The_Login_On_The_NuGetUser_Secret`, which reads the
`if:` line alone through `StepCondition`. So the gate was covered; what was wrong is that the cover was
**incidental**, living in a test named for a different question (#226's NUGET_USER distinction), where
narrowing that test to match its name would have dropped it with nothing failing. The issue is now
`priority:medium`, and the same overstatement was removed from the code comment #287 shipped.

### What changed

- **`TrustedPublishingWorkflowTests` reads comment-stripped text.** Its whole-slice assertions could not
  tell wiring from prose, and the login step's slice carries a 17-line comment block that names
  `release_created`. New test `ReleaseWorkflow_Login_Step_Should_Be_Read_Without_Its_Prose` pins that
  the slice has no comment line and names the gate token exactly once.
- **`Should_Gate_The_Login_On_A_Created_Release` now asserts on the `if:`, via `StepCondition`.**
  Stripping alone was not enough — a whole-slice `ShouldContain` over a *stripped* step is still
  satisfied by a non-gating mention. #302's *Alternatives* called this narrower fix "correct" and noted
  doing both is fine; both is what this is.
- **A duplicate step name fails loudly.** `TryStepNamed`/`StepNamed` count every match instead of taking
  the first — the shape `TryStepWithId` got in #267. Step ids are unique per job so a duplicate is
  invalid; step **names** are legal duplicates, so this was the looser primitive guarding the weaker
  property.
- **The name matcher accepts every legal spelling.** `StartsWith("- name: '…'")` missed double-quoted,
  unquoted and multi-space forms — so a duplicate spelled any of those ways still bound silently,
  leaving the new ambiguity guard easy to evade. Measured: `deploy-docs.yml` writes all ten of its step
  names unquoted, and the old matcher could not find a single one. The leading `-` stays **required**
  (unlike the id path, which may match a bare key and walk back) because every upload step carries
  `name:` inside its `with:` block as the artifact name.
- **Both scans now share `LinesMatching` + `ShouldNotBeAmbiguous`**, and both `Try*` forms take an
  optional `scopeDescription`; the two private `*OrNull` cores are gone.

### Code review

A review over `origin/dev...HEAD` raised 12 findings; **all 12 were addressed**. Three were substantive,
and each fix was verified against the reviewer's own scenario rather than assumed:

| Finding | Verified outcome |
|---|---|
| The ambiguity guard kept the naive matcher, so a duplicate in another legal spelling still bound silently — and `StepNamed` could not find any step in `deploy-docs.yml` | Tolerant matcher + 5 new tests, including the real `deploy-docs.yml` and a negative pin that `with: name:` must **not** answer for a step name |
| **Regression I introduced:** the loud ambiguity throws from inside `TryStepNamed`, which took no `scopeDescription` — so the sole production caller would have failed all five upload assertions with "the searched text …", re-creating the undifferentiated shared-reader failure #267 removed | `UploadStep` now passes `job.ToString()`. Measured by duplicating the upload step in `ci.yml`: the message now reads `ci.yml / build-and-test has 2 steps named 'Publish: test-results' …` |
| Stripping does not make the gate test "carry its own claim" — a non-gating `release_created` under `with:` still satisfies it | Routed through `StepCondition`. Measured with the `if:` deleted and `foo: ${{ … release_created }}` added under `with:`: the gate test now **fails**, where before it passed |

The other nine were accuracy and duplication findings, all correct and all mine — including a ⛔ comment
asserting the opposite of the paragraph two lines above it, a mis-attributed test reference for the
digest claim, three cross-references falsified by this very PR (they said
`TrustedPublishingWorkflowTests` hands over unstripped text — it no longer does), a present-tense
comment describing a defect this commit fixes, an unpinned `scopeDescription`, a missing whole-file-scope
caveat, a count assertion whose failure could not say which of two regressions it caught, and a third
near-copy of the match loop in a file whose class remark names duplication as the hazard it exists to
remove.

### Verification

- `./build.sh Test` (the exact CI target) — Restore/Compile/Test all succeeded, on the merged tree.
- `dotnet test -c Release` — **1559 passed, 1 skipped, 0 failed** (`FormCraft.UnitTests` 864, up from
  851 on `dev`; `ForMudBlazor` 537; `ForFluentUI` 158).
- `dev` synced twice during the run (5 sibling PRs landed); both merges clean, neither touched
  `.github/` or `build/Build.cs`, and the suite was re-verified on each merged tree.
- Workflows were edited only inside temporary experiments and restored; `git diff` shows **no change
  under `.github/`** in this PR.

## Follow-ups

- **A step written `-` alone with `name:` on the next line is not found by `StepNamed`.** Legal YAML, no
  workflow here writes it, and the miss is loud (reported absent) rather than silent. Supporting it means
  matching a bare `name:` key and walking back to the header — which is what the id path does, but
  `name:` collides with `with: name:` in every upload step, so it needs a containment rule rather than a
  looser regex. Deliberately out of scope here; documented at `StepNameKey`.
- **`TrustedPublishingWorkflowTests` is still file-scoped, not job-scoped** — tracked on #198. With
  duplicate *names* now failing loudly, a whole-file scope is slightly more exposed than before: two jobs
  each having a step of the same name is valid and would fail the scan. No workflow does this today, and
  the caveat is documented on `TryStepNamed`.


phmatray:
Noted while merging — **recorded, not filed** (no action planned):

A step written as `-` alone with `name:` on the following line is not found by `StepNamed`. It is legal YAML, but:
- **no workflow in this repo writes it** (all five use `- name:` on one line);
- the failure mode is **loud** — the step is reported absent and a test goes red — not the silent wrong-step binding #302 was about;
- the reason it is not supported is already documented in code, at `StepNameKey`: matching a bare `name:` key would require dropping the leading `-` anchor, and `name:` collides with the artifact name in every `with:` block, so it needs a containment rule (like the id path's walk-back) rather than a looser regex.

Deliberately not filed: it would commit someone to designing that containment rule for a shape nobody writes, and the code comment is the more durable place for the reasoning. This comment is the record if a workflow ever adopts that style.

PR #313, opened 2026-08-13
docs: correct CLAUDE.md's test commands for Microsoft.Testing.Platform (#299)

### LINKED ISSUE #299: CLAUDE.md still teaches VSTest test commands that Microsoft.Testing.Platform silently ignores
## Problem / motivation

`CLAUDE.md`'s **Running Tests** section is VSTest-era. The repository moved to
**Microsoft.Testing.Platform** (both `OutputType=Exe` + `UseMicrosoftTestingPlatformRunner=true`), and
MTP ignores VSTest options — so most of what that section teaches does nothing, **exits `0`, and says
so only in an `MTP0001` warning nobody reads**.

Measured on SDK `10.0.302`, xunit.v3 runner `3.2.2`, Release, macOS arm64:

| `CLAUDE.md` | Command | What actually happens |
|---|---|---|
| `:37-38` | `dotnet test --filter "FullyQualifiedName~FormBuilderTests"` | `MTP0001 … VSTestTestCaseFilter`; **the whole suite runs** (`Total: 808`) |
| `:41-43` | `dotnet test --filter "Category=Builder\|Renderer\|Security"` | same — *and* the repo has **zero `[Trait(...)]` attributes**, so these categories never existed |
| `:34` | `dotnet test --collect:"XPlat Code Coverage"` | `MTP0001 … VSTestCollect`; **no coverage file is produced at all**, exit `0` |
| `:26` | "600+ unit tests across 2 test projects" | **1340** tests across **three** projects |
| `:30-31` | lists 2 test projects | misses `FormCraft.ForFluentUI.UnitTests` (added in #261) |

Current counts: `FormCraft.UnitTests` 808 · `FormCraft.ForMudBlazor.UnitTests` 464 ·
`FormCraft.ForFluentUI.UnitTests` 68 = **1340**.

**Why this is worse than the same bug in the profile.** #277/#290 fixed exactly this class of error in
`.claude/skills/repo-profile.md` — but that file is read only by the `create-issue` / `implement-issue`
/ `merge-pr` lifecycle skills. **`CLAUDE.md` is auto-loaded into every session**, so any agent not
driving those skills still receives the disproven advice, and receives it silently: a filtered run that
quietly executes all 1340 tests looks like a fast pass, and a coverage run that emits no file looks
like success.

## Proposed solution

Correct the **Running Tests** section against the invocations #290 already measured and documented in
`.claude/skills/repo-profile.md` (the *Single-suite filter* bullet under *Build & test*):

1. Replace the five `--filter` examples with the working form —
   `dotnet test <csproj> -c Release -- --filter-class <FQN>` (everything after `--` reaches the MTP
   host), or the built host directly for the fast path.
2. Drop the `Category=` examples outright — there are no traits to match. Show class/namespace
   filtering instead.
3. Fix or remove the `--collect:"XPlat Code Coverage"` line, since it currently produces nothing.
4. Refresh the count ("1340 tests across 3 test projects") and add the missing FluentUI project.
5. Keep it **short and point at the profile** for the full trap list. #290 deliberately collapsed a
   duplicated copy of this guidance to stop the two drifting apart; re-duplicating it wholesale here
   would undo that. A correct example plus a pointer is the right size.

Additionally, add a small static guard under `FormCraft.UnitTests/Ci/` asserting `CLAUDE.md` contains
no `dotnet test --filter`/`--collect:` string — the same shape as the existing `Ci/GitignoreTests`
invariant, so the correction cannot silently rot back.

## Alternatives considered

- **Fix the prose only, no guard.** Smallest change. Rejected as the sole measure: this exact drift has
  now recurred three times (#231's inert VSTest wiring, #277's profile claim, this) and the failure mode
  is *silent*, so nothing surfaces it until someone re-measures by hand.
- **Delete the Running Tests section and point at the profile.** Tempting — one source of truth. Rejected:
  `CLAUDE.md` is the file every session actually loads, so it needs a usable command inline; a bare
  pointer to a lifecycle-skill config file is worse for the common reader.
- **A guard that shells out to the test host to prove the flags still work.** Rejected for the same
  reason #277 rejected it: circular (using the runner to test the runner's CLI), slow, and brittle
  against any xunit.v3 rename. The static text assertion has none of those properties.
- **Add trait attributes so the `Category=` examples become true.** Rejected: that is inventing test
  taxonomy to justify a doc line, and `--filter-trait` would still be the only way to use it.

## Area

Repository documentation / tooling — `CLAUDE.md`, test invocation, `FormCraft.UnitTests/Ci/`

---

**Follow-up from #290** (which corrected the same class of error in `.claude/skills/repo-profile.md`).
**Related:** #277, #261, #231, #223

<details>

### Problem / context

`CLAUDE.md` is the always-loaded briefing for any agent working in this repo. Its *Running Tests*
section predates the Microsoft.Testing.Platform migration, so it documents VSTest options that MTP
silently discards. Silence is the whole problem: every wrong command exits `0`, and the only signal is
an `MTP0001` build warning buried in restore output.

#277 established the true picture and #290 wrote it into `.claude/skills/repo-profile.md`. But that
file's own header scopes it to the lifecycle skills, so the correction reaches only agents driving
`create-issue`/`implement-issue`/`merge-pr`. The higher-traffic file was left stale — and it is the one
that shapes every other session's behaviour.

### Approaches

**A. Correct the prose, pointing at the profile for detail.** *Pros:* small, immediately removes the
wrong instructions, no new code, and respects #290's decision to keep the detailed trap list in exactly
one place. *Cons:* nothing stops a future edit from reintroducing a `--filter` example; the failure is
silent, so it would go unnoticed again.

**B. Correct the prose **and** add a static `Ci/` guard test.** A test asserting `CLAUDE.md` contains no
`dotnet test --filter`/`--collect:` invocation. *Pros:* the repo already has this exact pattern
(`FormCraft.UnitTests/Ci/GitignoreTests` pins a `.gitignore` invariant that had silently regressed and
was restored in #223); the assertion is a string check over a markdown file — fast, deterministic, no
toolchain dependency. *Cons:* one more test to update if the wording legitimately changes.

**C. Delete the section, defer entirely to the profile.** *Pros:* a single source of truth by
construction. *Cons:* `CLAUDE.md` is what actually gets loaded; removing runnable commands from it
degrades the common path to serve a tidiness goal.

### Recommendation

**B.** A alone repeats the mistake's own history — this is the third instance of VSTest-era wiring
lingering after the MTP move, and each was found only by someone re-measuring by hand. The guard is
cheap precisely because it is *static*: it does not run the test runner (which is what made the
analogous idea wrong in #277), it greps a documentation file, exactly like the `Ci/GitignoreTests`
precedent. C loses more than it gains.

</details>

<details>

### Goal

`CLAUDE.md`'s testing guidance is accurate under Microsoft.Testing.Platform, and a guard keeps it that
way.

### Scope

- Rewrite `CLAUDE.md`'s *Running Tests* block (roughly lines 24-44): filter examples, coverage line,
  test counts, project list.
- Add `FormCraft.UnitTests/Ci/ClaudeMdTestCommandsTests.cs` asserting the inert invocations are absent.

### Non-goals

- Changing `.claude/skills/repo-profile.md` again — #290 settled it; this issue points at it.
- Any change to build scripts, workflows, or `.csproj`. `./build.cmd Test` stays the CI gate and stays
  full-suite.
- Adding `[Trait]` attributes, or any test-taxonomy work.
- Auditing the rest of `CLAUDE.md` (its architecture/validation sections are out of scope).

### Behaviour

```mermaid
flowchart TD
    A["dotnet test --filter X"] --> B["MSBuild property VSTestTestCaseFilter"]
    B --> C["MTP0001 — ignored<br/>ALL 1340 TESTS RUN, exit 0"]
    D["dotnet test --collect:'XPlat Code Coverage'"] --> E["MTP0001 — VSTestCollect ignored<br/>NO coverage file, exit 0"]
    F["dotnet test &lt;csproj&gt; -c Release -- --filter-class X"] --> G["reaches the MTP host<br/>runs just that class"]
```

### Key files

- `CLAUDE.md` — the *Running Tests* section.
- `.claude/skills/repo-profile.md` — **read-only source** for the corrected wording (#290).
- `FormCraft.UnitTests/Ci/GitignoreTests.cs` — the shape the new guard copies.

### Validation rules

- No `dotnet test … --filter ` or `--collect:` invocation remains in `CLAUDE.md`.
- Every command written is run verbatim first and behaves as documented.
- Stated counts match a real run (currently 808 + 464 + 68 = 1340).
- The guard fails if a `--filter`/`--collect:` example is reintroduced.

### Edge cases

- The guard must not trip on prose *describing* the inert flags (this issue's own fix may want to warn
  "don't use `--filter`"). Assert on the **command form** (a line starting with `dotnet test` that
  contains `--filter`/`--collect:`), not on the bare substring.
- Counts drift as tests are added; keep the number approximate ("~1340") or accept it needs updating.

### Assumptions

- Both test projects keep the MTP runner; if either reverts to VSTest this guidance changes with it.
- Recorded from a `dev`-based checkout.

</details>

## 🛠️ Implementation plan

**Goal:** make `CLAUDE.md`'s testing guidance true under MTP, and guard it against silent regression.

**Architecture:** documentation plus one static invariant test. No library, build-script or workflow change.

**Tech stack:** .NET 10 SDK pinned by `global.json` (`10.0.302`), Microsoft.Testing.Platform + xunit.v3, xUnit + Shouldly.

**Global constraints:**
- ⛔ Do **not** change `build/Build.cs`, any `.github/workflows/*`, or any `.csproj`.
- ⛔ Do **not** re-edit `.claude/skills/repo-profile.md` — #290 owns that text; reuse it, don't duplicate it.
- `TreatWarningsAsErrors=true` is deliberate — do not relax it.
- Every documented command must be **run** before it is written down (this issue exists because unmeasured commands were).

### Task 1: Guard the invariant (failing test first)

**Files:** create `FormCraft.UnitTests/Ci/ClaudeMdTestCommandsTests.cs`; read `FormCraft.UnitTests/Ci/GitignoreTests.cs` for the file-locating pattern.

**Interfaces:** xUnit facts asserting no `dotnet test` line in `CLAUDE.md` carries `--filter ` or `--collect:`.

- [x] **Step 1:** Read `FormCraft.UnitTests/Ci/GitignoreTests.cs` and reuse its repo-root resolution so the new test locates `CLAUDE.md` the same way.
- [x] **Step 2:** Write the failing test — parse `CLAUDE.md` lines, select those starting with `dotnet test`, assert none contains `--filter ` or `--collect:`. Match the **command form**, not the bare substring, so prose warning about the flags is allowed.
- [x] **Step 3:** Run it → **FAIL**, listing the five `--filter` lines and the `--collect:` line. Record the failure output.
- [x] **Step 4:** Commit: `test(ci): pin CLAUDE.md against VSTest-era test invocations`.

### Task 2: Correct the Running Tests section

**Files:** modify `CLAUDE.md` (the *Running Tests* block, ~lines 24-44).

**Interfaces:** none.

- [x] **Step 1:** Measure first — run `dotnet test FormCraft.UnitTests/FormCraft.UnitTests.csproj -c Release -- --filter-class FormCraft.UnitTests.Ci.GitignoreTests` and confirm it runs a strict subset with a real `Passed!` line.
- [x] **Step 2:** Replace the `--filter` examples with that form, plus the direct-host fast path; note the host must be built first.
- [x] **Step 3:** Delete the three `Category=` examples (no `[Trait]` exists in the repo) and show class/namespace filtering instead.
- [x] **Step 4:** Fix or remove the `--collect:"XPlat Code Coverage"` line — measured, it emits `MTP0001 … VSTestCollect` and writes no coverage file.
- [x] **Step 5:** Update the count to 1340 across three projects and add `FormCraft.ForFluentUI.UnitTests` to the per-project list.
- [x] **Step 6:** Add a one-line pointer to `.claude/skills/repo-profile.md` (*Build & test* → *Single-suite filter*) for the full trap list — a pointer, not a copy, so the two cannot drift.
- [x] **Step 7:** Re-run Task 1's test → **PASS**.
- [x] **Step 8:** Commit: `docs: correct CLAUDE.md's test invocations for Microsoft.Testing.Platform`.

### Task 3: Verify end-to-end

**Files:** none modified.

**Interfaces:** none.

- [x] **Step 1:** Run every command now written in `CLAUDE.md` verbatim and confirm each behaves as documented (copy-paste errors in a doc of exact commands are the whole failure mode).
- [x] **Step 2:** Run `./build.sh Test` (CI gate) → all suites green under `TreatWarningsAsErrors`.
- [x] **Step 3:** Commit any fixes: `docs: correct a mis-transcribed command in CLAUDE.md` (skip if none).

phmatray:
Implements #299.

Closes #299.

`CLAUDE.md`'s **Running Tests** section is VSTest-era. The repo runs on Microsoft.Testing.Platform, which ignores VSTest options — so most of what that section teaches does nothing, exits `0`, and says so only in an `MTP0001` warning nobody reads. #290 fixed this class of error one level up in `.claude/skills/repo-profile.md`; this is the same fix in the file that is auto-loaded into *every* session.

code-review pass.

### Plan
- [x] Task 1: Guard the invariant (failing test first)
- [x] Task 2: Correct the Running Tests section
- [x] Task 3: Verify end-to-end

### What changed

`CLAUDE.md`'s *Running Tests* section now documents commands that actually do what they say, and a
static guard stops the old ones coming back.

| Was | Now |
|---|---|
| 5 × `dotnet test --filter …` | `dotnet test <csproj> -c Release -- --filter-class <FQN>`, plus the direct-host fast path (~200 ms) |
| `--filter "Category=Builder\|Renderer\|Security"` | removed — **no test in this repo carries a `[Trait]`**, so these never matched anything |
| `--collect:"XPlat Code Coverage"` | removed, with the reason stated: MTP discards `VSTestCollect` *and* no coverage extension is referenced, so no file was ever written |
| "600+ unit tests across 2 test projects" | "~1,550 across the three", deliberately approximate |
| 2 test projects listed | all three (FluentUI was missing) |

`FormCraft.UnitTests/Ci/ClaudeMdTestCommandsTests` fails the build if a `dotnet test` command in
`CLAUDE.md` regains `--filter`/`--collect`. It matches on the **command form**, not the bare
substring, so the corrected prose can still *name* the broken flags in order to warn about them — and
it folds `\`-continued lines first, since a wrapped command is the obvious way the guard would
otherwise be evaded by accident. `--filter-class` and friends deliberately still pass.

Red→green evidence: the guard failed on first run listing exactly the 6 offending commands, and
passes now. `./build.sh Test` is green — 1548 tests across the three projects.

### Code review

A review over `dev...HEAD` returned 12 findings; each was re-measured before acting. Ten were real
and are fixed in `fix: address code-review findings`. The ones that mattered:

- **The guard could go permanently vacuous.** Every assertion was of the form "no offender found",
  which is also what an *empty scan* reports — so a formatting change to the code fence would have
  made it pass forever while the broken commands sat unread. There is now an explicit
  `The_Scan_Should_Actually_Find_The_Documented_Commands` test. (The review also correctly showed my
  original companion test guarded a hole that cannot exist: `File.ReadAllLines` throws on a missing
  file, so it would never "silently pass". Replaced.)
- **The scan missed the path the doc recommends.** Anchoring on `dotnet test` left the direct-host
  invocation — the fast path CLAUDE.md now leads with, and so the likeliest place a future inert flag
  lands — completely unguarded. Verified by injecting
  `…/FormCraft.UnitTests --collect:"XPlat Code Coverage"`: the original guard passed, the new one
  fails with that exact line.
- **The verification rule was self-defeating.** "Require a real `Passed!`/`Failed!` line and a
  plausible count" is *satisfied* by the first `Passed! … Total: 6` of a solution-wide run whose other
  two assemblies reported `Failed! … Total: 0` with exit 1 — i.e. it green-lit the precise false pass
  this PR exists to prevent. Rewritten around per-assembly reporting and the unpiped exit code.
- **The mistyped-flag advice described only one path.** The `--help`/exit-5 behaviour is direct-host;
  via `dotnet test` the diagnostic never reaches stdout and you get `error run failed: Tests failed`
  pointing at a log — wording that sends you debugging source for an argument typo. Both documented.
- **"Fails the build" was simply wrong** — it is a unit test, so `dotnet test` catches it, not
  `dotnet build`. An agent validating with the build gate would have shipped the regression.
- **Debug/Release were mixed**: `dotnet test` with no `-c` builds Debug, then the documented Release
  host path does not exist. The block is uniformly Release now, matching CI.
- Also: MSBuild-property spellings (`-p:VSTestTestCaseFilter=…`) are now guarded and documented
  (measured — `MTP0001`, all 158 tests of the target project ran); a new test asserts every
  documented `--filter-class` FQN still resolves, so a rename cannot quietly turn the flagship
  example into a `Zero tests ran` no-op; Windows `.exe` caveat restored; flag list realigned.

**Scope call.** Two of my own changes falsified statements in `.claude/skills/repo-profile.md`, which
the plan lists as a non-goal to edit. It said *"`CLAUDE.md` still advertises `--filter "Category=Builder"`"*
— false as of this PR — and CLAUDE.md now names the profile authoritative, so leaving it would have
the two files contradicting each other. Its `808 + 464 + 68 = 1340` count had also gone stale within a
day (the real total is ~1,550). Both corrected minimally; not a broader profile audit.

**Dismissed:** the dangling `coverlet.collector` pin — real, but `.csproj`/props changes are forbidden
by the plan and it is already recorded below.

## Follow-ups

Adjacent findings, deliberately **not** fixed here (the plan forbids `.csproj`/props changes, and the
spec lists coverage work as out of scope):

- **`coverlet.collector` is a dead pin.** `Directory.Packages.props` carries a `PackageVersion` for it,
  but **no test project references it** and no CI workflow collects coverage — so the removed
  `--collect:` command could never have produced a file even on VSTest. Either wire coverage up
  (`Microsoft.Testing.Extensions.CodeCoverage` is the MTP route) or drop the pin.
- **There is currently no way to measure coverage in this repo.** Worth a decision either way; right
  now `CLAUDE.md` simply says so rather than offering a command that does nothing.

### DIFF
--- a/.claude/skills/repo-profile.md
+++ b/.claude/skills/repo-profile.md
@@ -59,9 +59,10 @@
       `Zero tests ran` (exit `8`), which is easy to skim past as success. A wildcard
       (`--filter-method '*Clearing_A_Standalone*'`) is the ergonomic escape.
     - ⚠️ **`--filter-trait` is dead weight here:** no test in this repo carries a `[Trait]`, so
-      `--filter-trait 'Category=Builder'` returns `Zero tests ran`. Note `CLAUDE.md` still
-      advertises `dotnet test --filter "Category=Builder"`, which is wrong twice over — the option
-      is inert *and* the trait does not exist. Filter by class or namespace instead.
+      `--filter-trait 'Category=Builder'` returns `Zero tests ran`. Filter by class or namespace
+      instead. (`CLAUDE.md` used to advertise `dotnet test --filter "Category=Builder"`, wrong twice
+      over — inert option, non-existent trait. Corrected in #299, which also added
+      `FormCraft.UnitTests/Ci/ClaudeMdTestCommandsTests` to stop it coming back.)
   - ⛔ **`dotnet test --filter …` really is inert** — the true half of the advice this bullet
     replaced. It is a VSTest option forwarded as an MSBuild property that MTP ignores, warning
     `MTP0001: VSTest-specific properties are set but will be ignored … VSTestTestCaseFilter` while
@@ -90,11 +91,13 @@
     and reports green on a command that failed — this single form defeats the rule directly above,
     which is why the example names its project.
   - **Scope:** this is *within-task* iteration only. It does **not** replace the *CI gates* below —
-    `./build.cmd Test` stays the pre-merge gate and still runs everything: 808 + 464 + 68 = 1340
-    tests, ~30s including `Compile`. Bare `dotnet test -c Release` runs the same three suites in
-    ~11s warm. ⚠️ **Neither prints a `1340` aggregate** — you get one `Passed!` line *per assembly*,
-    in nondeterministic order, so a check that reads the first one accepts `Total: 68` as the whole
-    suite. Confirm all three assemblies reported, or trust the process exit code for the full run
+    `./build.cmd Test` stays the pre-merge gate and still runs everything: **~1,550 tests** across
+    the three projects, ~30s including `Compile`. Bare `dotnet test -c Release` runs the same three
+    suites in ~11s warm. (Approximate on purpose — the precise `808 + 464 + 68 = 1340` recorded here
+    by #290 was stale within a day, which is the whole failure mode this section is about.)
+    ⚠️ **Neither prints an aggregate** — you get one `Passed!` line *per assembly*,
+    in nondeterministic order, so a check that reads the first one accepts one project's total as the
+    whole suite. Confirm all three assemblies reported, or trust the process exit code for the full run
     (unpiped). Filter to iterate; run one of these before you claim done.
 - **Format/lint apply:** `dotnet format FormCraft.sln`
 - **Format/lint verify (the gate):** *CI runs no `dotnet format` check.* The enforcing gate is the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -29,12 +29,14 @@ see the warning below the commands, which is the part that costs time if you ski
 ```bash
 # Run everything — ~1,550 tests across the three test projects (approximate on purpose: the
 # exact figure drifts with every merge, and a stale precise number reads as authoritative)
-dotnet test
+dotnet test -c Release
 
-# Run one test project
-dotnet test FormCraft.UnitTests/FormCraft.UnitTests.csproj
-dotnet test FormCraft.ForMudBlazor.UnitTests/FormCraft.ForMudBlazor.UnitTests.csproj
-dotnet test FormCraft.ForFluentUI.UnitTests/FormCraft.ForFluentUI.UnitTests.csproj
+# Run one test project. `-c Release` throughout: CI runs Release, and the host path
+# below is a Release path — mixing configurations is how you end up running one build
+# and inspecting another.
+dotnet test FormCraft.UnitTests/FormCraft.UnitTests.csproj -c Release
+dotnet test FormCraft.ForMudBlazor.UnitTests/FormCraft.ForMudBlazor.UnitTests.csproj -c Release
+dotnet test FormCraft.ForFluentUI.UnitTests/FormCraft.ForFluentUI.UnitTests.csproj -c Release
 
 # Run one class — everything after `--` is forwarded to the test host.
 # Always name the .csproj: run solution-wide, the filter is applied to all three
@@ -43,14 +45,16 @@ dotnet test FormCraft.UnitTests/FormCraft.UnitTests.csproj -c Release \
   -- --filter-class FormCraft.UnitTests.Ci.GitignoreTests
 
 # Same filter without the build step (~200ms). Run `dotnet build -c Release` first,
-# or you are testing a stale binary.
+# or you are testing a stale binary. On Windows the host is `...\FormCraft.UnitTests.exe`.
 FormCraft.UnitTests/bin/Release/net10.0/FormCraft.UnitTests \
   --filter-class FormCraft.UnitTests.Ci.GitignoreTests
 ```
 
-Filters: `--filter-class`, `--filter-method`, `--filter-namespace`, their `--filter-not-*`
-counterparts, and `--filter-uid`. `*` wildcards work at either end. `--filter-method` wants the
-**fully-qualified** name (`<namespace>.<class>.<method>`) — a bare method name matches nothing.
+Filters: `--filter-class`, `--filter-method`, `--filter-namespace`, `--filter-trait`, `--filter-uid`,
+`--filter-query`, plus a `--filter-not-*` counterpart for class/method/namespace/trait. `*` wildcards
+work at either end, and the simple filters cannot be combined with `--filter-query`.
+`--filter-method` wants the **fully-qualified** name (`<namespace>.<class>.<method>`) — a bare method
+name matches nothing. `--filter-trait` is useless here: no test in this repo carries a `[Trait]`.
 
 ⛔ **The VSTest spellings are silently ignored here.** Passing `--filter` to `dotnet test` (rather
 than after `--`) forwards it as an MSBuild property that Microsoft.Testing.Platform discards with a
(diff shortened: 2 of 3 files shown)


phmatray:
Follow-up triage for this merge (per `/merge-pr` Step 6):

Both items in this PR's Follow-ups section share one root cause — coverage is unwired, and a vestigial
package pin implies otherwise — so they were clustered and filed as a single issue rather than two:

- **#326** — *The `coverlet.collector` pin has no consumer, implying a coverage setup that does not exist.* The "no way to measure coverage" item is captured there as the alternative (wire up `Microsoft.Testing.Extensions.CodeCoverage`), which the issue deliberately keeps separate from the dead-pin removal so a cleanup does not smuggle in a capability decision.

Nothing else was deferred in the PR body or review comments. For the record, the `.worktrees/` item
raised back on #290 is already tracked as **#316**, so it was not refiled here.

Issue #310, opened 2026-08-13
Clean deletes the build output the running Nuke process is executing from


phmatray:
## Problem / motivation

Since #275 (PR #288, `27dcaa1`), `Clean` sweeps every project in the solution:

```csharp
Solution.AllProjects
    .Select(project => project.Directory)
    .SelectMany(directory => directory.GlobDirectories("**/bin", "**/obj"))
    .ToList()
    .ForEach(directory => directory.DeleteDirectory());
```

`build/_build.csproj` is in `FormCraft.sln`, so this deletes `build/bin` and `build/obj` — **the
directory the running Nuke process is executing from**. Both wrappers launch it from there:

| Wrapper | Line | Command |
|---|---|---|
| `build.sh` | 67 | `"$DOTNET_EXE" run --project "$BUILD_PROJECT_FILE" --no-build -- "$@"` |
| `build.ps1` | 73-74 | `dotnet build $BuildProjectFile` then `dotnet run --project $BuildProjectFile --no-build -- $BuildArguments` |

So the live assembly is `build/bin/Debug/net10.0/_build.dll`, loaded through its apphost, and `Clean`
removes the directory containing it while it runs.

**Measured on macOS arm64** (`dotnet --version` = `10.0.302`): `./build.sh Clean` succeeds repeatedly,
exit `0`, and `build/bin` + `build/obj` are gone afterwards. POSIX permits unlinking an open file, so
the running image survives in memory and nothing complains.

**Not measured on Windows, and that is the concern.** A loaded PE image is locked by the OS, so
`Directory.Delete` over `build/bin` is expected to throw `UnauthorizedAccessException`/`IOException`
and fail the target. `./build.ps1` is a documented entry point in `CLAUDE.md`, so this would be a
plain `./build.ps1 Clean` failing for Windows contributors.

**CI cannot catch it.** `ci.yml` and `continuous.yml` run on `ubuntu-latest` only, and neither invokes
`Clean` at all — they run `./build.cmd Test` and `./build.cmd Pack`, and `Clean` is `.Before(Restore)`
without being a dependency of anything. The target only ever runs when a human asks for it, which is
precisely the population most likely to be on Windows.

There is a smaller cost on **every** platform, already visible today:

- A chained invocation (`./build.sh Clean Compile`) deletes the running process's own not-yet-loaded
  assemblies mid-run, so any lazily-resolved Nuke/Serilog type needed by a later target can throw
  `FileNotFoundException`.
[...]

PR #328, opened 2026-08-13
chore(deps): drop the unused coverlet.collector pin (#326)

### LINKED ISSUE #326: The coverlet.collector pin has no consumer, implying a coverage setup that does not exist
## Problem / motivation

Three statements about code coverage in this repo disagree with each other, and the one a reader is
most likely to believe is the false one.

| Where | Says | Reality |
|---|---|---|
| `Directory.Packages.props:35` | `<PackageVersion Include="coverlet.collector" Version="10.0.1" />` | **no project references it** — not `FormCraft.UnitTests`, `FormCraft.ForMudBlazor.UnitTests`, `FormCraft.ForFluentUI.UnitTests`, nor `Directory.Build.props` |
| `build/Build.cs`, `.github/workflows/ci.yml` | *(nothing)* | no CI step collects coverage |
| `CLAUDE.md` (until #313) | `dotnet test --collect:"XPlat Code Coverage"` | produced **no coverage file at all**, exit `0` |

The third was removed in #313. The first is still there, and it is the one that implies the capability
exists.

**Measured** (SDK `10.0.302`, Release): `dotnet test FormCraft.ForFluentUI.UnitTests/… --collect:"XPlat
Code Coverage"` emits `warning MTP0001: VSTest-specific properties are set but will be ignored … 
VSTestCollect`, runs all 158 tests, exits `0`, and writes no `.cobertura.xml` — or any other coverage
artifact — anywhere under the project. That is expected: the test projects are Microsoft.Testing.Platform
hosts, and MTP discards the VSTest data collector that `--collect:` drives.

**Why the dead pin is not harmless.** Renovate reads `Directory.Packages.props`, so it keeps proposing
upgrades for a package nothing consumes (it appears on the Dependency Dashboard, #4) — recurring review
noise for a no-op. And a contributor who greps for "coverage", finds the pin, and concludes coverage is
configured will reach for exactly the command #313 just deleted.

## Proposed solution

**Remove the `coverlet.collector` pin**, leaving one consistent story: this repo does not measure
coverage, and nothing claims it does.

`CLAUDE.md` already states this outright (#313), so after removing the pin all three places agree.
Restore was never possible through that package anyway — under MTP the route is
`Microsoft.Testing.Extensions.CodeCoverage` (`dotnet test -- --coverage`), which is a different
package and a deliberate choice nobody has made.

If coverage *is* wanted, that is the alternative below — but it is a capability decision, not a
cleanup, and it should not ride along on a dead-pin removal.

## Alternatives considered

- **Wire coverage up properly:** reference `Microsoft.Testing.Extensions.CodeCoverage` in the three
  test projects and emit a report from `./build.cmd Test`. *Pros:* a 1,500-test suite with no coverage
  signal is a real gap. *Cons:* it adds a CI artifact, a runtime cost and a threshold nobody has asked
  for; it is scope creep on "this pin has no consumer". Left as a follow-on if the owner wants it —
  removing the pin does not block it, since the MTP package is a different dependency anyway.
- **Leave the pin and document it as unused.** *Pros:* zero risk. *Cons:* a comment explaining why a
  dependency is dead is strictly worse than not having the dead dependency; Renovate still files PRs
  for it.
- **Add a `Ci/` guard asserting every `PackageVersion` has a consumer.** Tempting — the repo has the
  idiom (`Ci/GitignoreTests`, and #302/#303 are about exactly this family). Rejected here as
  disproportionate to a `priority:low` chore, and the general invariant deserves its own issue rather
  than being smuggled in behind one dead pin.

## Area

Build/dependency configuration — `Directory.Packages.props`, test tooling

---

**Follow-up from #313** (which removed the coverage command from `CLAUDE.md`).
**Related:** #299, #4, #276, #301

<details>

### Problem / context

#277 → #290 → #299 → #313 form a chain about one theme: this repository migrated to
Microsoft.Testing.Platform, and the VSTest-era wiring left behind fails **silently and green**. #290
corrected the repo profile, #313 corrected `CLAUDE.md` and pinned it with a guard test. This issue is
the last visible residue of that migration: the coverage package that the deleted command would have
used.

It is a small thing, but it is the *kind* of small thing this chain exists to stamp out — a
configuration that asserts a capability the toolchain cannot deliver, where nothing fails loudly enough
for anyone to notice.

### Approaches

**A. Remove the pin.** *Pros:* one-line change; makes `Directory.Packages.props`, CI and `CLAUDE.md`
agree; stops the Renovate churn; carries no risk because nothing consumes it. *Cons:* leaves the repo
with no coverage measurement — but that is already true today, so nothing is lost.

**B. Wire up MTP coverage.** *Pros:* actually delivers the capability the pin implies; a test suite
this size arguably warrants it. *Cons:* new dependency, new CI artifact, slower runs, and an implicit
promise to keep a threshold honest. That is a product decision, and bundling it into the removal of an
unrelated dead pin would make both harder to review.

**C. Keep the pin, add an explanatory comment.** *Pros:* nothing moves. *Cons:* documents a defect
instead of fixing it, and Renovate does not read comments.

### Recommendation

**A.** It is the only option that is purely a *correction* — it removes a false claim without adding a
commitment. B remains open afterwards and is genuinely cheaper from a clean base: the MTP extension is a
different package, so nothing about removing `coverlet.collector` has to be undone to adopt it.

</details>

<details>

### Goal

Nothing in the repository implies coverage is configured, because nothing is.

### Scope

- Remove the `coverlet.collector` `<PackageVersion>` entry from `Directory.Packages.props`.
- Confirm restore/build/test are unaffected and no documentation still promises coverage.

### Non-goals

- Adding coverage collection (that is approach B, and a separate decision).
- Changing `build/Build.cs`, any workflow, or any `.csproj`.
- Re-editing `CLAUDE.md`'s Running Tests block or `.claude/skills/repo-profile.md` — #313 settled both.
- A general "every PackageVersion has a consumer" guard — worth its own issue, not this one.

### Behaviour

```mermaid
flowchart TD
    A["Directory.Packages.props<br/>pins coverlet.collector"] --> B["no .csproj references it"]
    B --> C["Renovate still proposes upgrades<br/>reader infers coverage exists"]
    D["dotnet test --collect:'XPlat Code Coverage'"] --> E["MTP0001 — VSTestCollect ignored<br/>NO file written, exit 0"]
    F["remove the pin"] --> G["props, CI and CLAUDE.md<br/>all agree: no coverage"]
```

### Key files

- `Directory.Packages.props` — the only file changed (line 35).
- `CLAUDE.md` — already states coverage is unwired (#313); read-only here.

### Validation rules

- No `coverlet` reference remains anywhere in the repository.
- `dotnet restore` and `dotnet build -c Release` succeed with **0 warnings** under
  `TreatWarningsAsErrors`.
- `./build.cmd Test` stays green (~1,550 tests across the three projects).

### Edge cases

- **Central Package Management:** removing a `PackageVersion` breaks the build only if some project
  carries a matching `<PackageReference Include="coverlet.collector" />`. Verified: none does — so the
  removal is inert. Re-verify with a clean `dotnet restore` rather than assuming.
- A stale `obj/` can mask a restore problem; prefer a clean restore when checking.

### Assumptions

- The test projects keep the MTP runner. If one reverted to VSTest, `coverlet.collector` would become
  meaningful again and this removal should be revisited.

</details>

## 🛠️ Implementation plan

**Goal:** remove the dead `coverlet.collector` pin so nothing implies coverage is configured.

**Architecture:** build configuration only — one line of `Directory.Packages.props`.

**Tech stack:** .NET 10 SDK pinned by `global.json` (`10.0.302`), Central Package Management, Microsoft.Testing.Platform + xunit.v3.

**Global constraints:**
- ⛔ Do **not** add coverage collection — that is the rejected-here alternative B.
- ⛔ Do **not** change `build/Build.cs`, any `.github/workflows/*`, or any `.csproj`.
- `TreatWarningsAsErrors=true` is deliberate — do not relax it.
- Verify by **running** the commands; this chain of issues exists because unmeasured claims were written down.

### Task 1: Prove the pin has no consumer, then remove it

**Files:** modify `Directory.Packages.props` (remove the `coverlet.collector` `<PackageVersion>` line).

**Interfaces:** none.

- [x] **Step 1:** Prove it is unreferenced before touching anything: `grep -rn "coverlet" --include='*.csproj' --include='*.props' --include='*.targets' .` — expect the single `Directory.Packages.props` hit and no `PackageReference`. Record the output.
- [x] **Step 2:** Remove the `<PackageVersion Include="coverlet.collector" ... />` line.
- [x] **Step 3:** Run a clean restore: `dotnet restore` → succeeds (a missing `PackageVersion` for a referenced package fails restore under Central Package Management, so this is the real check, not the build).
- [x] **Step 4:** Run `dotnet build -c Release` → **0 warnings, 0 errors** under `TreatWarningsAsErrors`.
- [x] **Step 5:** Commit: `chore(deps): drop the unused coverlet.collector pin`.

### Task 2: Confirm nothing still promises coverage

**Files:** none modified — verification only.

**Interfaces:** none.

- [x] **Step 1:** `grep -rn "coverlet\|XPlat Code Coverage\|--collect" --include='*.md' --include='*.cs' --include='*.yml' --include='*.props' .` — expect no hit that *offers* coverage (a hit explaining that coverage is unwired, in `CLAUDE.md` or the repo profile, is correct and stays).
- [x] **Step 2:** Run `./build.sh Test` (the CI gate) → all three suites green, ~1,550 tests.
- [x] **Step 3:** Commit any doc correction found in Step 1: `docs: drop a stale coverage reference` (skip if none).

phmatray:
Implements #326.

Closes #326.

`Directory.Packages.props` pins `coverlet.collector`, but no project references it and no workflow collects coverage — so the pin's only effects are recurring Renovate upgrade PRs for a package nothing consumes, and implying to a reader that coverage is configured. #313 already removed the `--collect:"XPlat Code Coverage"` command it would have served (measured: `MTP0001`, no coverage file written, exit `0`). This removes the last piece.

code-review pass.

### Plan
- [x] Task 1: Prove the pin has no consumer, then remove it
- [x] Task 2: Confirm nothing still promises coverage

### Verification

- **The pin had no consumer.** `grep -rn "coverlet" --include='*.csproj' --include='*.props' --include='*.targets' .` returned exactly one hit — the `PackageVersion` itself. No `PackageReference` anywhere.
- **`dotnet restore` succeeds** after removal. That is the real check under Central Package Management: a removed `PackageVersion` for a *referenced* package fails restore outright, so a clean restore proves the pin was inert.
- **`dotnet build -c Release`** → 0 warnings, 0 errors under `TreatWarningsAsErrors`.
- **`./build.sh Test`** (the CI gate) → green, 1553 tests across the three projects.
- **Nothing still offers coverage.** The remaining `coverlet`/`--collect` matches are all correct to keep: `CHANGELOG.md` (release-please-owned history, never hand-edited), and `CLAUDE.md` + `Ci/ClaudeMdTestCommandsTests` doc comments, which *warn* that the flag is inert rather than offering it. No doc correction was needed, so Task 2 Step 3 committed nothing — as the plan allows.

Coverage remains unmeasured in this repo, which is now stated consistently in all three places instead of being contradicted by a pin. Wiring it up (`Microsoft.Testing.Extensions.CodeCoverage`) stays available and is unaffected by this removal — it is a different package.

### Code review

**Clean for this PR — no fix commit.** The reviewer's verdict on the change itself: *"the pin drop
itself is clean (no `.csproj` still references `coverlet.collector`, and CLAUDE.md was already updated
to say `--collect:"XPlat Code Coverage"` writes nothing here). The solution builds with 0 warnings/0
errors."*

⚠️ **The review over-scoped, and that was my error.** I ran it as `dev...HEAD`, but the *local* `dev`
ref in this worktree was stale (`1ea5425`) while `origin/dev` had advanced many merges beyond it — so
the three-dot range resolved against the stale ref and the reviewer read **15 commits / ~10.4k inserted
lines** of already-merged sibling work instead of my one-line diff. Verified afterwards:
`git diff --stat origin/dev...HEAD` → `1 file changed, 1 deletion(-)`.

Its 12 findings are therefore **not about this PR** and are deliberately **not fixed here** — they
concern code already on `dev`. They looked substantive rather than noise, so they are recorded below
rather than discarded.

### Merge sync

Merged `origin/dev` in (a large sibling batch: demo refactor, `FieldConfigurationTracker`,
`WorkflowSource` rework). **No conflicts** — nothing else touched `Directory.Packages.props`. Re-verified
on the merged tree rather than trusting the clean text merge: pin still absent with no new consumer,
`dotnet restore` clean, `./build.sh Test` green at **1572 tests**.

## Follow-ups

⚠️ These came from the **over-scoped review described above** and concern **already-merged code, not
this PR**. Recorded so they are not lost; each needs confirming against `dev` before being actioned,
and they should be triaged on their own merits rather than attributed to a one-line dependency change.

**Already tracked — do not refile:**
- *`Clean` deletes the bin/obj the running Nuke process was loaded from* (Windows `IOException`; CI is
  ubuntu-only so it never fires there) — this is **#310**.

**Fluent UI adapter parity gaps** (from #278/#291's new adapter) — these cluster as one root: the new
adapter reimplements behaviours MudBlazor already had, and several diverge:
- Multi-file upload throws `ArgumentException` when the model property is `List<IBrowserFile>` —
  `GetMultipleFiles` returns `BrowserFile[]`; the component's own tests pass a `List<>` so they cannot
  catch it.
- `FluentUILovFieldComponent.ResolveSelectionValue` can only ever publish one value in multi-select
  (`values is TValue` is true only when `TValue` is `object`), so the UI shows N chips while the model
  holds one.
- LOV `SelectRowAsync` fires `OnDependencyChanged` *before* publishing its own value, so dependent
  fields recompute from the stale value (MudBlazor does the reverse).
- LOV/Lookup never re-read the model after `OnInitialized` and seed display text from
  `CurrentValue.ToString()` rather than the display selector — a preloaded model renders no chips.
- Lookup `SelectRowAsync` silently returns on a null/unmatched selector result: the row click does
  nothing, panel stays open, no diagnostic.
- `FluentUIFileUploadComponentBase` reads `MaxFileSize` as `long`, so a `.WithAttribute("MaxFileSize", 5_000_000)`
  `int` is silently discarded and the 10 MB default applies.
- Autocomplete option-provider path can throw `TargetException`/`InvalidCastException` out of the
  keystroke handler instead of degrading.
- Collection field: reorder/remove leaves index-keyed validation messages attached to the old position.

**Other, independent:**
- `MudBlazorFileUploadFieldComponent` never reads `FileUploadConfiguration`, so `.AsFileUpload(acceptedFileTypes:…, maxFileSize:…)`
  is ignored on the single-file path — the multi-file sibling does read it. (Same *class* as the Fluent
  `MaxFileSize` item above.)
- Lookup/LOV grids put `role="button"` on `<tr>`, destroying table semantics, and handle Space in
  `@onkeydown` without `preventDefault`, so selecting also scrolls the page.
- `FormCraft.DemoFluentApp/wwwroot/index.html` ships the default `maximum-scale=1.0, user-scalable=no`
  viewport — blocks pinch-zoom (WCAG 2.1 1.4.4), in a repo that holds itself to Level A.


phmatray:
Follow-up triage for this merge (per `/merge-pr` Step 6).

⚠️ **Context that matters for how these were routed:** this PR's Follow-ups came from a **mis-scoped
code review** — it ran as `dev...HEAD` against a stale local `dev` ref, so it read ~10.4k lines of
already-merged sibling work rather than this PR's one-line diff. The findings are therefore about code
already on `dev`, not about this change. I refused to file unverified bug reports against other
people's merged work, so each item below was either **verified first**, **found to be already tracked**,
or **recorded rather than filed**.

### Filed (verified before filing)

- **#340** — *Upload components read their constraints ad-hoc, so `AsFileUpload` is silently ignored on
  the single-file MudBlazor path* (`type:bug`, `priority:medium`). Verified directly: `AsFileUpload`
  writes **only** `WithAttribute("FileUploadConfiguration", config)`, while
  `MudBlazorFileUploadFieldComponent` reads only the raw `"Accept"`/`"MaxFileSize"`/`"MaxFiles"` keys and
  never that configuration — so every constraint from the public builder API is dropped there. The
  multiple-file sibling *does* read it, so the two disagree. The comment in `AsFileUpload` asserting
  "the component… reads this configuration attribute" is false for that component. Clustered with the
  reported Fluent `int`/`long` coercion bug, since both are instances of one root: four components each
  resolving constraints their own way.
- **#341** — *The Fluent UI demo's viewport disables pinch-zoom (WCAG 2.1 1.4.4)* (`type:bug`,
  `priority:low`). Verified: `DemoFluentApp/wwwroot/index.html:6` carries
  `maximum-scale=1.0, user-scalable=no`; `DemoBlazorApp/wwwroot/index.html:6` carries neither — the
  newer demo regressed against the reference by inheriting the stock template line.

### Already tracked — not refiled

- *`Clean` deletes the bin/obj the running Nuke process was loaded from* → **#310**.
- *LOV/Lookup never re-read the model after `OnInitialized`* → closely owned by **#335** (the Fluent
  adapter caching field configuration at init); worth confirming there rather than opening a second
  issue on the same "initialises once, never re-syncs" root.
- *Collection reorder/remove leaves index-keyed validation messages on the wrong row* → adjacent to
  **#334** (collection rows losing component identity when reordered/removed) and **#329**.

### Recorded, not filed — unverified by me

These looked substantive but I did not confirm them, and they concern merged code that other work is
already touching. Listing them so they are retrievable; each needs checking against `dev` before anyone
acts on it:

- `FluentUIMultipleFileUploadComponent`: a model property declared `List<IBrowserFile>` may throw
  `ArgumentException`, because `GetMultipleFiles` returns `BrowserFile[]` — reportedly invisible to the
  component's own tests, which pass a `List<>` in directly.
- `FluentUILovFieldComponent.ResolveSelectionValue`: reportedly can only publish one value in
  multi-select (`values is TValue` holds only when `TValue` is `object`), so the UI would show N chips
  while the model holds one.
- `FluentUILovFieldComponent.SelectRowAsync`: reportedly fires `OnDependencyChanged` *before* publishing
  its own value, so dependent fields recompute from the stale value (MudBlazor does the reverse).
- `FluentUILookupFieldComponent.SelectRowAsync`: reportedly returns silently on a null/unmatched
  selector result — row click does nothing, panel stays open, no diagnostic.
- `FluentUIAutocompleteFieldComponent`: reportedly can throw `TargetException`/`InvalidCastException`
  out of the keystroke handler instead of degrading.
- Lookup/LOV grids: `role="button"` on `<tr>` (destroying table semantics) and Space handled in
  `@onkeydown` without `preventDefault` (so selecting also scrolls the page).

No deferred work originated from this PR's own change — it was a one-line dependency removal.

Issue #276, opened 2026-08-12
The .html test report is promised but unenforced, so half the artifact can vanish silently


phmatray:
## Problem / motivation

The `Test` target promises three artifact kinds and enforces one.

Since #256 (PR #259) the target declares, per test project:

```csharp
.Produces(TestResultsDirectory / "**/*.trx")
.Produces(TestResultsDirectory / "**/*.html")
.Produces(TestResultsDirectory / "**/*.log")
```

and after the run it asserts — per project, naming the offender — that a `*.trx` landed in that
project's `test-results/<project>/`. Nothing asserts the `*.html`.

So if a future xunit.v3 or Microsoft.Testing.Platform release drops, renames or stops honouring
`--report-xunit-html` while keeping `--report-xunit-trx`, every existing check still passes:

- the per-project trx guard passes — the trx is still written;
- `BuildScript_Should_Emit_Reports_Through_The_Testing_Platform` passes — it asserts the **flag is
  present in `Build.cs`'s text**, not that a file appeared;
- `Test_Target_Should_Only_Promise_Artifacts_The_Runner_Is_Asked_To_Write` passes — same reason, it
  cross-checks the promise against the *flags*, not against the output;
- the three workflows upload under `if-no-files-found: ignore`, so an absent html raises nothing.

The html half of the artifact disappears and no build, test or workflow anywhere says a word. That is
exactly the defect #231 was filed about — "declared but never emitted", surviving months because
nothing checked — at half scale, in the same file, guarded against on one axis and not the other.

The html is not redundant with the trx, which is what makes the gap worth closing rather than
accepting: the trx is the machine-readable record, the html is the one a human opens from a downloaded
artifact without tooling. #225 and #231 exist to serve exactly that reader.

## Proposed solution

Extend the guard already in the loop. `Test` iterates the test projects to check the trx; ask the same
question about the html in the same pass, and name the project in the same way:

```csharp
foreach (var kind in new[] { "*.trx", "*.html" })
    // …assert per completed project, message naming project + directory + the missing kind
```

Better still, derive the checked set from the `.Produces` declarations rather than restating it, so the
promise and the guard cannot drift — which is the whole failure mode.

## Alternatives considered

- **Leave it; the trx is the important one.** Defensible until the day someone downloads an artifact
  to read the html and finds it silently absent. The guard costs two lines and closes the class, not
  the instance.
[...]

PR #333, opened 2026-08-13
test(ci): cover the missing-upload-step branch with a synthetic job set (#303)

### LINKED ISSUE #303: The Ci guard suite can only run against the real workflows, so its absence paths are untested
## Problem / motivation

`TestReportingTests` guards five claims about the `Publish: test-results` upload step, and **none of
their absence branches can execute in a normal run**.

The guard set is discovered once, statically, from the repository's own workflows:

```csharp
private static readonly IReadOnlyList<TestRunningJob> TestRunningJobs = DiscoverJobsThatRunTests();
// -> WorkflowSource.Matching(InvocationPattern) -> WorkflowSource.JobsOf(workflow)
```

Every job that reaches Nuke's `Test` target in `.github/workflows/` **does** have the upload step — that
is what the suite exists to keep true — so `UploadStep(job)` never returns `null`, and the absence half
of each assertion is dead code in every run:

```csharp
private static bool UploadStepFails(TestRunningJob job, Func<string, bool> claim) =>
    UploadStep(job) is not { } step || !claim(step);
//  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ never true when CI is green
```

That branch is the behaviour change at the heart of #267 (PR #287): a job missing the step used to throw
from inside the shared helper, reddening four tests with `Array.FindIndex(…) should be greater than or
equal to 0 but was -1` — an apparent fault in the reader rather than in the workflow — and now each test
reports the offending `workflow / job` pair itself. **Writing `is { } step &&` instead of
`is not { } step ||` would ship green**, and a missing step would read as *satisfying* every claim.

#287 did verify it, by hand, exactly as #267's plan prescribed: delete the step from `ci.yml`, confirm
all five tests name `ci.yml / build-and-test` as an offender, revert. That proof is real, and it is not
repeatable — it lives in a PR description, not in CI.

The contrast inside that one PR is the tell: the `WorkflowSource` half gained eight fixture-driven tests
because it takes its scope **as a parameter**, while this half gained none because its input is welded to
the repository. Same PR, same author, same discipline — the difference is purely testability.

This is the shape #231 was filed about, one level up: wiring that reads correct and is never exercised.
Here the unexercised wiring is in the *guard*, which is worse, because a guard is what you consult
instead of looking.

## Proposed solution

Give the five assertions their job set as an argument, defaulting to the discovered one, so a synthetic
job can be fed in.

Extract the offender computation:

```csharp
private static List<TestRunningJob> UploadOffenders(
    IReadOnlyList<TestRunningJob> jobs,
    Func<string, bool> claim) =>
    jobs.Where(j => UploadStepFails(j, claim)).ToList();
```

Each existing `[Fact]` then reads `UploadOffenders(JobsThatRunTests(), <claim>).ShouldBeEmpty()` —
unchanged in meaning — and one new test runs **all five claims** against a hand-built job whose text has
no upload step, asserting each reports it. The vacuity guard stays in `JobsThatRunTests()`, and the
`(workflow, job)` offender labelling comes from `TestRunningJob.ToString()` either way.

## Alternatives considered

- **A fixture-based overload of `UploadStep` only.** Smaller, and it would let a test drive the
  `TryStepNamed` call — but the thing at risk is not the lookup (already covered by
  `WorkflowSourceTests`), it is the five `||` branches that decide whether absence counts as an
  offender. Only injecting the *job set* reaches those.
- **Leave it; the manual check was done.** It was, and it found nothing wrong. But it cannot run again
  without a human remembering to, which is the same standing this file's own remarks call out about
  #231: "nothing anywhere noticed months of empty output". A guard is exactly the thing that should not
  depend on someone remembering.
- **Assert the absence path by temporarily mutating a real workflow inside the test.** Writing to
  `.github/workflows/` from a test is unacceptable — it races the other suites reading the same files
  (`WorkflowSource` caches them in static state) and can leave the repo dirty on a failed run.
- **Make `WorkflowSource.All` injectable instead.** Would let a test substitute the whole workflow set,
  and is a much bigger change: `RepoRoot`, `All`, `BuildScript` and the strip cache are all static
  singletons deliberately (read-once over files that cannot change during a run). Reopen only if a
  second suite needs synthetic workflows.

## Area

Test infrastructure — `FormCraft.UnitTests/Ci/TestReportingTests.cs`. No library code, no workflow
files, no `build/Build.cs` change, no new package.

**Related:** #267 (landed as PR #287; added the absence branches and verified them manually), #255
(extracted the shared reader `WorkflowSource`), #256 (per-project results directories), #231 and #252
(why these upload guards exist at all), #276 (a sibling gap in the `Test` target's artifact guard — a
promise that is declared but never verified against output).

<details>

### Problem / context

`FormCraft.UnitTests/Ci/` asserts on repository files because what it guards runs a few times a year or
fails silently. That design has one structural cost: the suite's inputs are the repo itself, so any
branch that only fires on a *broken* repo cannot be reached from a *healthy* one.

#267 added exactly such branches — five of them — and then had to prove them by breaking `ci.yml` by
hand. The proof worked. The problem is that it was a one-off: nothing stops a later refactor from
inverting one of those conditions, and the suite would stay green while a job silently stopped being
checked for `if: always()`.

`WorkflowSource` does not have this problem, because #255 gave it text-taking primitives
(`StepNamed(scope, …)`, `JobsOf(fileName)`), and `WorkflowSourceTests` drives them with inline YAML
fixtures. The asymmetry is not philosophical; it is one parameter.

### Approaches

**A. Inject the job set into the five assertions.** Extract `UploadOffenders(jobs, claim)`; the `[Fact]`s
pass `JobsThatRunTests()`, a new test passes a synthetic list.
*Trade-offs:* smallest change that actually reaches the five `||` branches. Each existing test keeps its
name, its claim and its message. Adds one indirection to a file that already has `UploadStep` and
`UploadStepFails` helpers, so it fits the grain.

**B. Overload `UploadStep` with a text-taking form.** Lets a test exercise the lookup on synthetic text.
*Trade-offs:* does not reach the decision under test. The lookup is already covered by
`WorkflowSourceTests`' `TryStepNamed` tests; what is uncovered is whether each *assertion* treats `null`
as an offender. Rejected as not addressing the gap.

**C. Make `WorkflowSource.All` injectable.** Substitute the whole workflow set for a test.
*Trade-offs:* the most general, and it would serve `TrustedPublishingWorkflowTests` too. But the static
read-once caches are a deliberate design (`RepoRoot`, `All`, `StrippedCache`, `BuildScript`), and
unpicking them for one suite's benefit trades a real property for a hypothetical one. Revisit if a second
suite needs it.

### Recommendation

**A.** It reaches the exact branches that shipped unverified, in one file, without touching the reader's
caching design. B misses the target; C is a bigger change than the defect justifies.

**Assumptions:** the five existing tests keep their current names and messages, so a failure reads the
same to anyone who has seen one before. `TestRunningJob` is a private nested record, so tests in the same
class can construct one without widening its visibility. The suite stays green throughout — this adds
coverage rather than changing a claim.

</details>

<details>

### Goal

Every absence branch in `TestReportingTests`' upload assertions is exercised by an automated test, so an
inverted condition fails the build instead of shipping green.

### Scope

- `FormCraft.UnitTests/Ci/TestReportingTests.cs` — extract `UploadOffenders(jobs, claim)`, route the five
  `[Fact]`s through it, add the synthetic-job coverage.

### Non-goals

- No new NuGet dependency; no YAML parser (#267's rejection stands and its reopening criterion —
  assertions turning materially structural — is not met here).
- No change to any workflow, to `build/Build.cs`, or to library code. In particular, **no test writes to
  `.github/workflows/`**.
- Not making `WorkflowSource.All`/`RepoRoot` injectable (approach C).
- Not changing what any of the five claims asserts, nor their failure messages.

### Surface

```
private static List<TestRunningJob> UploadOffenders(
    IReadOnlyList<TestRunningJob> jobs, Func<string, bool> claim)
```

Private, and the five `[Fact]`s become one line each. `UploadStep` and `UploadStepFails` keep their
current signatures; `UploadOffenders` is the seam.

### The claim table

The five claims, and what each must report for a job with no upload step:

| Test | Claim over the step | Absence must be |
|---|---|---|
| `Every_Job_That_Runs_Tests_Should_Upload_The_TestResults_Artifact` | the step exists | an offender |
| `Every_TestResults_Upload_Should_Run_On_Failure` | contains `if: always()` | an offender |
| `Every_TestResults_Upload_Should_Point_At_The_Directory_The_Build_Fills` | a line trimming to `path: test-results` | an offender |
| `No_TestResults_Upload_Should_Point_At_A_Path_The_Build_No_Longer_Fills` | does **not** contain `**/TestResults/` | an offender |
| `Every_TestResults_Upload_Should_Tolerate_A_Missing_Path` | contains `if-no-files-found: ignore` | an offender |

The fourth is the one to keep straight: its claim is inverted, so it reads "the step exists **and**
avoids the stale glob". Absence offends it deliberately — a missing step makes that test red today (by
throwing), and letting it read as vacuously satisfied would trade a loud failure for a quieter suite.

### Edge cases

- A synthetic job whose text is empty: `TryStepNamed` returns `null`, so every claim reports it. Same
  path as "no upload step", which is the intent.
- The vacuity guard lives in `JobsThatRunTests()` and must stay there — the new test passes its own list
  and so bypasses it, which is correct: a hand-built list cannot be accidentally empty.
- A synthetic job that *does* have a conforming step must be reported by **none** of the claims — worth
  asserting in the same test, so the new coverage cannot pass by reporting everything.

### Assumptions

`TestRunningJob(string Workflow, string Job, string Text)` stays constructible from within the class.
Its `ToString()` (`"$"{Workflow} / {Job}"`) is what makes the offender legible, so the synthetic job is
given realistic values (`"fixture.yml"`, `"build"`) rather than placeholders.

</details>

## 🛠️ Implementation plan

**Goal:** every absence branch in the upload assertions is exercised automatically, so an inverted
condition fails the build.

repository files rather than on library behaviour, which is why the seam has to be introduced
deliberately.

**Global Constraints**

- `TreatWarningsAsErrors=true` is deliberate — never relax it to make a build pass.
- **No new NuGet package**, and no YAML parser (#267 closed that question).
- No change to any workflow, to `build/Build.cs`, or to library code. **No test may write to
  `.github/workflows/`** — `WorkflowSource` caches those files in static state, so a mutating test
  would race every other suite in the class.
- Keep the five tests' names, claims and failure messages as they are — a failure should read the same
  to someone who has seen one before.

### Task 1: Introduce the job-set seam and cover the absence path

**Files:** modify `FormCraft.UnitTests/Ci/TestReportingTests.cs`.

**Interfaces:** adds `private static List<TestRunningJob> UploadOffenders(IReadOnlyList<TestRunningJob> jobs, Func<string, bool> claim)`.
`UploadStep(TestRunningJob) -> string?` and `UploadStepFails(TestRunningJob, Func<string,bool>) -> bool`
keep their signatures.

- [x] **Step 1:** Write the failing test in `TestReportingTests.cs`. It drives all five claims over two
      synthetic jobs — one with no upload step, one with a conforming step — so the coverage cannot pass
      by reporting everything:

```csharp
[Fact]
public void Every_Upload_Claim_Should_Report_A_Job_That_Has_No_Upload_Step()
{
    // The branch this pins shipped in #267 verified only by hand (deleting the step from ci.yml, then
    // reverting). It cannot fire in a normal run, because every real test-running job HAS the step —
    // so writing `is { } step &&` instead of `is not { } step ||` would ship green, and a missing
    // upload would read as satisfying all five claims.
    var missing = new TestRunningJob("fixture.yml", "build", """
                                                             steps:
                                                               - name: 'Run: Test'
                                                                 run: ./build.sh Test
                                                             """);

    var conforming = new TestRunningJob("fixture.yml", "good", $$"""
                                                               steps:
                                                                 - name: '{{UploadStepName}}'
                                                                   if: always()
                                                                   uses: actions/upload-artifact@v4
                                                                   with:
                                                                     path: test-results
                                                                     if-no-files-found: ignore
                                                               """);

    foreach (var claim in UploadClaims)
    {
        UploadOffenders([missing], claim.Value).ShouldBe([missing], $"the '{claim.Key}' claim does not report a job with no upload step");
        UploadOffenders([conforming], claim.Value).ShouldBeEmpty($"the '{claim.Key}' claim reports a conforming step as an offender");
    }
}

/// <summary>
/// The five claims this file makes about the upload step, keyed by the test that owns each — so the
/// absence coverage above cannot drift out of step with the assertions themselves.
/// </summary>
private static Dictionary<string, Func<string, bool>> UploadClaims => new(StringComparer.Ordinal)
{
    ["exists"] = _ => true,
    ["if: always()"] = s => s.Contains("if: always()", StringComparison.Ordinal),
    ["path: test-results"] = s => s.Split('\n').Any(line => line.Trim() == "path: test-results"),
    ["avoids **/TestResults/"] = s => !s.Contains("**/TestResults/", StringComparison.Ordinal),
    ["if-no-files-found: ignore"] = s => s.Contains("if-no-files-found: ignore", StringComparison.Ordinal),
};
```

- [x] **Step 2:** Run `dotnet test -c Release` → **RED** (`UploadOffenders` does not exist).
- [x] **Step 3:** Add the seam next to the existing helpers:

```csharp
/// <summary>
/// The jobs in <paramref name="jobs" /> whose upload step fails <paramref name="claim" /> — including
/// by not existing.
/// </summary>
/// <remarks>
/// Takes the job set rather than reading <see cref="JobsThatRunTests" /> directly, which is the only
/// way the absence branch can be reached: every real test-running job has the step, so a test that
/// cannot supply its own jobs cannot exercise "the step is missing" at all.
/// </remarks>
private static List<TestRunningJob> UploadOffenders(
    IReadOnlyList<TestRunningJob> jobs,
    Func<string, bool> claim) =>
    jobs.Where(j => UploadStepFails(j, claim)).ToList();
```

- [x] **Step 4:** Run `dotnet test -c Release` → **GREEN** for the new test.
- [x] **Step 5:** Route the five existing `[Fact]`s through the seam, leaving each claim and message
      exactly as it is. For example:

```csharp
var offenders = UploadOffenders(
    JobsThatRunTests(),
    s => s.Contains("if: always()", StringComparison.Ordinal));

offenders.ShouldBeEmpty();
```

      and the inverted one:

```csharp
var offenders = UploadOffenders(
    JobsThatRunTests(),
    s => !s.Contains("**/TestResults/", StringComparison.Ordinal));

offenders.ShouldBeEmpty();
```

- [x] **Step 6:** Run `dotnet test -c Release` → **GREEN**, with the pass count up by exactly one (the
      new test) and no existing test renamed or removed.
- [x] **Step 7:** Prove the coverage bites: temporarily invert one claim's branch in `UploadStepFails`
      (`is { } step &&`), confirm `Every_Upload_Claim_Should_Report_A_Job_That_Has_No_Upload_Step` turns
      **red**, then restore it. This is the check #267 could only run by editing a workflow.
- [x] **Step 8:** Update the `UploadStep`/`UploadStepFails` remarks to record that the absence rule is
      now covered by a synthetic job set rather than by the manual `ci.yml` deletion described in #267.
- [x] **Step 9:** Commit: `test(ci): cover the missing-upload-step branch with a synthetic job set`

phmatray:
Implements #303.

Closes #303.

Gives `TestReportingTests` a seam so its own assertions can be exercised against synthetic jobs. Test
infrastructure only — one file. No library code, no workflow files, no `build/Build.cs` change, no new
package.

### Plan
- [x] Task 1: Introduce the job-set seam and cover the absence path

### What changed

`TestReportingTests` discovers its guard set once, statically, from the repository's real workflows.
Every real test-running job **has** the `Publish: test-results` upload step — that is what the suite
exists to keep true — so the five assertions' "the step is missing" branch could never execute in a
green run. #267 verified it by hand (delete the step from `ci.yml`, watch the tests name the job,
revert), which proved the behaviour once and could not prove it again.

- **`UploadOffenders(jobs, claim)`** takes the job set instead of reading `JobsThatRunTests()` itself.
  That parameter is the whole point: a test that cannot supply its own jobs cannot reach the branch.
- **Each content claim is now defined once** — `RunsOnFailure`,
  `PointsAtTheDirectoryTheBuildFills`, `AvoidsThePathTheBuildNoLongerFills`, `ToleratesAMissingPath` —
  referenced both by the `[Fact]` that holds it against the real workflows and by the coverage
  theories. A dictionary of re-typed copies would let the two drift apart silently.
- **`UploadMissing(jobs)`** names the presence check, so the test whose subject is "the step is there"
  no longer reads as *jobs that fail an always-true claim*.
- **Three coverage tests**: every claim reports a job with no step; every claim reports a step that is
  *present and violates it*; and a conforming step offends nothing.

### Code review

A review raised 11 findings; **all 11 addressed**. One was decisive and two more were real gaps in the
coverage I had written — the first draft of this PR was half-built:

| Finding | Why it mattered |
|---|---|
| **No fixture had the step present but violating a claim** | A `UploadStepFails` that ignored `claim` entirely would have passed every test, silently reducing all four content assertions to "the step exists" — the exact vacuity #303 was filed about, one branch over |
| **Both fixtures were passed as single-element lists** | An implementation returning the *whole list* whenever any job offended would have passed — and in production (3 jobs) that destroys the "names the offending job" property this file is built around |
| **`UploadClaims` duplicated the predicates** rather than being what the `[Fact]`s read, while its doc claimed the opposite guarantee | 165 lines of separation between the copies; drift would have been invisible in both directions |
| The `conforming` fixture omitted `name: test-results` and pinned `@v4` | That `name:`/`path:` pair is the one discrimination the path claim exists to make, so the fixture could not exercise it. Now copied from `ci.yml` verbatim |
| A `foreach` over claims in one `[Fact]` aborts at the first failure | Now `[Theory]`, so each claim reports independently — the form the sibling `WorkflowSourceTests` already uses |

The rest were accuracy fixes, again mostly in my prose: a doc promising a guarantee nothing enforced,
"drives every claim over a synthetic job" (the delegate is never invoked on that path — `||`
short-circuits), and "cannot fire in a normal run" said of a branch that fires exactly when the guard
does its job.

### Verification

The coverage was **mutation-tested**, not assumed. Three mutations, each of which a reviewer identified
as previously shipping green:

| Mutation | Before | After |
|---|---|---|
| `UploadStepFails` ignores `claim` entirely | passed | **4 tests fail** |
| Absence branch inverted (`is { } step &&`) | caught | **4 tests fail** |
| `UploadOffenders` returns the whole list if any job offends | passed | **8 tests fail** |

All restored afterwards; the tree is clean.

- `./build.sh Test` (the exact CI target) — Restore/Compile/Test all succeeded on the merged tree.
- `dotnet test -c Release` — **1601 passed, 1 skipped, 0 failed** (`FormCraft.UnitTests` 876;
  `ForMudBlazor` 567; `ForFluentUI` 158).
- **`dev` moved three times during the run.** The last merge **overlapped this file** — #276 (PR #325)
  also edited `TestReportingTests.cs`. Git merged it without conflict, and I checked the result was
  *semantically* clean rather than trusting that: #276's additions are confined to the `Build.cs`
  report-kind assertions, touch nothing in the upload region, and all four production call sites still
  route through the named delegates.


phmatray:
**Merge blocked — not by this PR.** `dev` is red, and merging it in makes this PR's CI red too.

- This PR's own CI was **green** at `62225e4` (build-and-test, ubuntu-latest, lint-pr-title all passed).
- Syncing `dev` in (`87d84d2`) brought in the failure: `CollectionItemShapeGuardTests.No_Suite_Should_Re_Declare_A_Collection_Item_Shape_The_Fixture_Provides`, in `FormCraft.ForMudBlazor.UnitTests` — a project this PR does not touch.
- Reproduced on a **pristine checkout of `origin/dev`** with none of this branch's commits: `Failed: 1, Passed: 582`. `dev` has been red on every commit since `0ad0d49`.

Cause and remedies are in #345 — a collision between #306 (added the guard) and #308 (added the model it rejects), each green on its own branch.

Holding this PR rather than merging over a red bar. Once `dev` is green, re-run `/merge-pr #333`: the branch is already synced, reviewed and verified, so it should land without further work.

phmatray:
**Unblocked.** `dev` is green again — #346 (`a68cf29`) fixed the shape-guard collision that #345 tracked, and #345 is closed.

Re-synced this branch onto `dev` (`8d23d45`) and re-verified, including a gate that did **not** exist when this PR was written: #307 landed `ci(build): enforce dotnet format`, so the build now runs `dotnet format --verify-no-changes` over the whole solution.

- `./build.sh Test` — **1625 passed, 1 skipped, 0 failed** (884 / 584 / 158)
- `./build.sh Format` (the new gate) — **exit 0**, no violations in this branch

Ready to land.

